PR #86, opened 2022-03-31
fix: correct flame script

GaoxiangLuo:
Corrected the line numbers discrepancy caused by the addition of a license.

type | file | changes
-- | -- | --
modified | fiab/flame.sh | changed line numbers


myungjin (approved):
lgtm

PR #95, opened 2022-04-06
fix: fedavg algo

myungjin:
The previous global weights were not reset before new global weights
are computed in each round. This caused loss value increase over
rounds. This error is fixed.


GaoxiangLuo:
lgtm

GaoxiangLuo (approved):

GaoxiangLuo (approved):
lgtm

PR #97, opened 2022-04-08
take load_data function out of loop

myungjin:
It's enough to call load_data function only once.
Also, sometimes parameters derived from the loaded dataset needs to be
fed to model architecture initialization.

To accomodate these aspects, load_data function is moved before
initialization task.

In doing so, a bug in composer is identified. The bug skips the first
tasklet in a loop from a second iteration. The bug is also fixed.


GaoxiangLuo (approved):
lgtm

PR #98, opened 2022-04-09
refactor internal logic

myungjin:
internal logic is updated as follows.

1) user doesn't need to udpate model or weights across rounds any
more in his job script. The modules in mode/horizontal take care of
updating model and weights automatically.

2) trainer can publish its metrics to mlflow registry.


GaoxiangLuo (approved):
lgtm. From now on, the local models are registered to MLFlow and the evaluation step is moved from the central server to the local clients as the central server doesn't have the access to data.

PR #104, opened 2022-04-14
features: add fedopt aggregation

GaoxiangLuo:
Added three adaptive aggregation algorithms FedAdam, FedAdaGrad, and FedYogi to the aggregator. This implementation was done in PyTorch, and Tensorflow version will be supported later.

type | file | changes
-- | -- | --
modified | lib/python/flame/config.py | rewrote new initialization for optimizer; registered more optimizer options
modified | lib/python/flame/examples/mnist/aggregator/config.json | example config
modified | lib/python/flame/examples/mnist/trainer/config.json | example config
modified | lib/python/flame/mode/horizontal/middle_aggregator.py | passed parameters into optimizers
modified | lib/python/flame/mode/horizontal/top_aggregator.py | passed parameters into optimizers
created | lib/python/flame/optimizer/fedopt.py | abstract class of fedopt - adaptive optimizer family
created | lib/python/flame/optimizer/fedadagrad.py | passed parameters into optimizers
created | lib/python/flame/optimizer/fedadam.py | passed parameters into optimizers
created | lib/python/flame/optimizer/fedyogi.py | passed parameters into optimizers
modified | lib/python/flame/optimizers.py | registered more optimizers

### DIFF
--- a/lib/python/flame/optimizer/fedadagrad.py
+++ b/lib/python/flame/optimizer/fedadagrad.py
@@ -1,0 +1,39 @@
+# Copyright 2022 Cisco Systems, Inc. and its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""FedAdaGrad optimizer"""
+"""https://arxiv.org/abs/2003.00295"""
+import logging
+
+from .fedopt import FedOPT
+
+logger = logging.getLogger(__name__)
+
+class FedAdaGrad(FedOPT):
+    """FedAdam class."""
+    logger.debug("Initializing fedadagrad")
+
+    def __init__(self, beta_1=0.9, beta_2=0.99, eta=1e-2, tau=1e-3):
+        """Initialize Fedadagrad instance."""
+        super().__init__(beta_1, beta_2, eta, tau)
+
+    def _delta_v_pytorch(self):
+        self.v_t = {k: self.v_t[k] + self.d_t[k]**2 for k in self.v_t.keys()}
+        return
+
+    def _delta_v_tensorflow(self):
+        # TODO: tensorflow implementation
+        raise NotImplementedError("Tensorflow implementation not yet implemented")
--- a/lib/python/flame/optimizer/fedadam.py
+++ b/lib/python/flame/optimizer/fedadam.py
@@ -1,0 +1,39 @@
+# Copyright 2022 Cisco Systems, Inc. and its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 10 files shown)


lib/python/flame/optimizer/fedadagrad.py
```diff
@@ -0,0 +1,39 @@
...
+        """Initialize Fedadagrad instance."""
+        super().__init__(beta_1, beta_2, eta, tau)
+
+    def _delta_v_pytorch(self):
+        self.v_t = {k: self.v_t[k] + self.d_t[k]**2 for k in self.v_t.keys()}
+        return
+
+    def _delta_v_tensorflow(self):
+        # TODO: tensorflow implementation
+        pass
```
myungjin, line 39: Perhaps better to raise NotImplementedError since any tensorflow experiment will do nothing?

If there is no side-effect, it may be okay to leave as is.

Please consider how to handle this.

lib/python/flame/optimizer/fedadam.py
```diff
@@ -0,0 +1,39 @@
...
+        """Initialize FedAdam instance."""
+        super().__init__(beta_1, beta_2, eta, tau)
+
+    def _delta_v_pytorch(self):
+        self.v_t = {k: self.beta_2 * self.v_t[k] + (1 - self.beta_2) * self.d_t[k]**2 for k in self.v_t.keys()}
+        return
+
+    def _delta_v_tensorflow(self):
+        # TODO: tensorflow implementation
+        pass
```
myungjin, line 39: Same here

lib/python/flame/optimizer/fedyogi.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""FedYogi optimizer"""
+"""https://arxiv.org/abs/2003.00295"""
+import logging
+
+from .fedopt import FedOPT
+
+import torch
```
myungjin, line 23: Can you move this to line 36? This will break tensorflow implementation of FedYogi when tensorflow version is implemented.

lib/python/flame/optimizer/fedopt.py
```diff
@@ -0,0 +1,103 @@
...
+    def do(self, cache: Cache, total: int):
+        """Do aggregates models of trainers.
+
+        Return: aggregated model
+        """
+        logger.debug("calling fedopt")
+
+        self.agg_weights = super().do(cache, total)
+        if self.agg_weights is None:
+            return None
```
myungjin, line 67: Should `current_weights` need to be returned here so that at least the latest weights can be used for the remaining rounds?

lib/python/flame/optimizer/fedopt.py
```diff
@@ -0,0 +1,103 @@
...
+from abc import abstractmethod
+import logging
+
+from diskcache import Cache
+
+from .fedavg import FedAvg
+from ..common.util import (MLFramework, get_ml_framework_in_use,
+                           valid_frameworks)
+
+import torch
```
myungjin, line 28: can you move this to line 84 due to the same reason in the comment left for FedYogi?

myungjin (commented):
Thank you for this PR. I left a couple of comments.

Did you also do some benchmark testing for each of the algorithms?

lib/python/flame/optimizer/fedadagrad.py
GaoxiangLuo: Okay, I will raise a NotImplementedError. Also, the benchmarking testing will come as a separate PR with the documentation of optimizers.

lib/python/flame/optimizer/fedopt.py
GaoxiangLuo: That makes sense to me. If there are no trainers, which causes `agg_weights` to be None this round, we can still return the `current_weights`.

myungjin (approved):
lgtm

PR #105, opened 2022-04-18
refactor: rename MedNIST to MedMNIST

GaoxiangLuo:
Change the Python library example MedNIST to MedMNIST -- name changes only.


myungjin (approved):
lgtm

PR #106, opened 2022-04-19
feature: parallel group training

GaoxiangLuo:
This is a system-level example to demonstrate how to execute parallel group training.

type | file | changes
-- | -- | --
created | examples/parallel_group/dataset_asian_china.json | example dataset
created | examples/parallel_group/dataset_eu_uk.json | example dataset
created | examples/parallel_group/dataset_us_west.json | example dataset
created | examples/parallel_group/job.json | example job
created | examples/parallel_group/mnist.zip | example code
created | examples/parallel_group/schema.json | example schema
modified | lib/python/flame/mode/horizontal/top_aggregator.py | customized run name in MLFLow
modified | lib/python/flame/mode/horizontal/trainer.py | customized run name in MLFLow
modified | lib/python/flame/registry/mlflow.py | reset the experiment when trying to set an existed experiment


lib/python/flame/registry/mlflow.py
```diff
@@ -48,7 +48,11 @@ def __new__(cls):
     def __call__(self, uri: str, experiment: str) -> None:
         """Initialize the instance."""
         mlflow.set_tracking_uri(uri)
-        mlflow.set_experiment(experiment)
+        try:
+            mlflow.set_experiment(experiment)
+        except:
```
myungjin, line 53: Can you catch the exact error or error value?

You mentioned that there was MlflowException with RESOURCE_ALREADY_EXISTS error code, or something similar.

Can you please add more precise error handling code?

lib/python/flame/mode/horizontal/trainer.py
```diff
@@ -55,8 +55,14 @@ def internal_init(self) -> None:
...
-        name = self.config.role + '-' + self.config.agent_id[:8]
-        self.registry_client.setup_run(name)
+        groupby_value = ""
+        for v in self.config.channels.values():
+            for val in v.groupby.value:
+                if val in self.config.realm:
+                    groupby_value = groupby_value + val + "-" 
+
+        run_name = self.config.role + '-' + groupby_value + self.config.agent_id[:8]
+        self.registry_client.setup_run(run_name)
```
myungjin, line 65: This code block for composing run_name with groupby can be used in various places (e.g., distributed learning).

Even in this PR, the same code is repeated twice.

It would be better to make it as a function and place the function in some common place (e.g., `flame/common/util.py` or creating util.py under `flame/mode` looks okay too).

myungjin (commented):
left a few comments.

Can you please also add documentation on the parallel experiment?

The example name can be parallel_experiments instead of parallel group as parallel group doesn't mean much.

lib/python/flame/mode/horizontal/trainer.py
GaoxiangLuo: good idea

lib/python/flame/registry/mlflow.py
GaoxiangLuo: Yes, I can catch the exact error - when two roles tried to `set_experiment` and they both found the experiment hadn't been created, so they both tried to create a new experiment with the same experiment ID. However, there is always one role faster than the other in creating the experiment, causing the experiment_already_exist error. 

Solution: re-call the `set_experiment` again does consistently solve the issue, but I will add a couple of more log messages.

myungjin (approved):
lgtm

PR #107, opened 2022-04-20
implement adult income example

myungjin:
Using adult census dataset, income prediction example is implemented
as a federated learning job. The example only has a single trainer for
the sake of simplicity; more trainers can be added.


lib/python/setup.py
```diff
@@ -13,37 +13,18 @@
...
-    version='0.0.7',
-    author='Myungjin Lee',
-    author_email='myungjle@cisco.com',
+    version='0.0.8',
+    author='Flame Maintainers',
+    author_email='flame-github-owners@cisco.com',
     include_package_data=True,
     packages=find_packages(),
-    # TODO: remove data_files later as it is not essential
-    data_files=[('flame/examples/hier_mnist/top_aggregator',
```
GaoxiangLuo, line 29: Do we really want to remove these files? Would it be less intuitive for users to test the examples in the manual mode with the python library?

GaoxiangLuo: Actually, now `lib/python/MANIFEST.in` will do the same thing, but maybe consider `*.json` instead of `config.json` as some configuration files have slightly different naming such as `config_uk.json` in hier_mnist example.

myungjin: MANIFEST.in will take care of adding all config files in the examples folder. To add all .json files recursively, I will change `config.json` to `*.json`.

GaoxiangLuo (approved):
lgtm

PR #115, opened 2022-04-23
support optimizer option in control plane

myungjin:
Flame library supports several optimizers such as FedAvg, FedAdaGrad,
FedAdam, and FedYogi. To make the control plane use these options,
Flame openapi specification is updated to include optimizer options.


GaoxiangLuo (approved):
lgtm

PR #116, opened 2022-04-26
task log report

myungjin:
When task(s) gets failed, it is hard to debug since no log is recorded
in the system. Now a last 1000 byte text from the logs of the task is
obtained and recorded in the system.

Also, agentId is renamed as taskId since taskId makes more sense than
agentId. This caused refactoring throughout the entire repo.

Some of deprecated cdoe and example are removed too.

MLflow exception is properly handled in registry/mlflow.py.


GaoxiangLuo (approved):
lgtm

PR #117, opened 2022-04-27
feature: distributed training with mqtt

GaoxiangLuo:
This is an example of distributed training with the MQTT backend. It's for demonstration purposes so It doesn't have efficient communication backends for now. This distributed training mode will work even if there is only one trainer. The newly-joined trainer will receive the latest aggregated weights before start contributing.

type | file | changes
-- | -- | --
modified | lib/python/flame/config.py | add the number of distributed learning nodes to the configuration
modified | lib/python/flame/mode/message.py | add new message types
created | lib/python/flame/examples/dist_mnist/\__init__.py | 
created | lib/python/flame/examples/dist_mnist/trainer/\__init__.py | 
created | lib/python/flame/examples/dist_mnist/trainer/config.json | example config
created | lib/python/flame/examples/dist_mnist/trainer/main.py | borrowed from MNIST example
created | lib/python/flame/mode/distributed/\__init__.py | 
created | lib/python/flame/mode/distributed/trainer.py | created a new trainer class to handle distributed mode

### DIFF
--- a/lib/python/flame/config.py
+++ b/lib/python/flame/config.py
@@ -66,7 +66,7 @@ GROUPBY_DEFAULT_GROUP = 'default'
 DEFAULT_HYPERARAMETERS_DICT = {
     'rounds': 1,
     'epochs': 1,
-    'batchSize': 16,
+    'batchSize': 16
 }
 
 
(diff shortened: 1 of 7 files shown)


myungjin (commented):
left two minor comments.
Also, please verify if aggregated models are identical across workers on every round (e.g., by computing a hash value of aggregated model weights).

GaoxiangLuo:
I've made sure the aggregated model weights across trainers on every round are identical.

myungjin (approved):
lgtm

PR #123, opened 2022-05-10
fix: deadlock case in default selector

myungjin:
A deadlock occurs when workers join late after aggregator distributes
global parameters. This is because the aggregator will wait for local
parameters from those workers to which it didn't distribute the global
parameters. To solve the issue, the default selector remembers a set
of workers it distributed the global parameters in each round and
within the round the aggregator will try to receive local parameters
from the same set of workers.


GaoxiangLuo (approved):
lgtm

PR #126, opened 2022-05-11
feature: per-channel backend support

myungjin:
Each channel can have its own backend instead of using a single
backend. This allows per-channel optimization.


GaoxiangLuo (approved):
lgtm

PR #131, opened 2022-05-18
feature: hybrid mode with ring allreduce

GaoxiangLuo:
type | file | changes
-- | -- | --
modified | examples/distributed_training/dataset_1.json | changed realm to be the same to run distributed training
modified | examples/distributed_training/dataset_2.json | changed realm to be the same to run distributed training
modified | examples/distributed_training/dataset_3.json | changed realm to be the same to run distributed training
modified | examples/distributed_training/job.json | 
modified | examples/distributed_training/schema.json | modified schema to run distributed training with flame system control
-- | -- | --
created | lib/python/flame/examples/hybrid/trainer/config_eu_org1.json | rename
created | lib/python/flame/examples/hybrid/trainer/config_eu_org2.json | rename
created | lib/python/flame/examples/hybrid/trainer/config_us_org1.json | rename
created | lib/python/flame/examples/hybrid/trainer/config_us_org2.json | rename
-- | -- | --
modified | lib/python/flame/mode/distributed/trainer.py | example config
modified | lib/python/flame/mode/hybrid/trainer.py | major change of this commit

This commit adapt the ring all-reduce method from distributed training to the hybrid mode as hybrid mode uses distributed training among the grouped trainers. 

There is one issue that needs to be resolved in this PR. 
1) The system works if there is one trainer and one aggregator.
2) The system works if there are two trainers from two different groups respectively with one aggregator.
3) The system works as expected when two trainers of the same group start first, and then start the aggregator.
4) The system fails when starting with one trainer and one aggregator, and then start another trainer of the same group (i.e., a new trainer joins in the midst of training).

There is one issue that needs to be resolved in future PR for distributed learning example.
1) When there are 2 trainers already forming 2-node ring topology and close to the end of training, a new trainer joins and cannot form an agreement of the 3-node ring topology. It's possible that the new trainer is selected as one of the nodes saving the local model, which isn't what we want. This case won't occur in the hybrid mode because a new trainer in hybrid mode will fetch the latest weights from the aggregator first.
2) Currently, a number of selected trainers picked by COMMIT-COUNT will be the representative sending the weights to the aggregator. However, this isn't necessary as after ring all-reduce operation all trainers should share the same weights. They should still select a trainer to send the weights instead of multiple.


myungjin:
Closing this PR since PR #170 is created.

PR #133, opened 2022-05-21
feature: ingress support

myungjin:
As a step towards multi-cluster support, control plane
components (apiserver and notifier) need to be exposed  to an external
world. To achieve that, kubernetes ingress controller is enabled.
Along the way, dns support is also needed. Since it is hard (doesn't
make sense) to obtain a public domain name for a fiab environment,
a local dns resolver service (ingress-dns, a minikube's addon) is
enabled and configured.

In addition, tls termination is enabled in the nginx ingress
controller because notifier uses grpc and the nginx ingress controller
can only support grpc when both tls and http2 are enabled.

The domain used for testing (flame.test) is a private domain name. So,
selfsigned certificates are created to use ingress and ingress-dns.

For mlflow, tls is not enabled as mlflow complains about a failure of
verifying the selfsigned certficate and it is not easy to enable an
insecure connection as the feature is buried deep in mlflow's code.

flame's control plane code is also updated to support tls and insecure
connection.

Documents are updated accordingly to explain these changes.


GaoxiangLuo (commented):
This commit belongs to the modification of control plane of source code in go language. Pass it to Dhruv.

GaoxiangLuo (approved):
lgtm.
1. MLFlow server changes host domain.
2. `--insecure` flag to `flamectl`
3. broker and mlflow has prefix `flame-` in config.

PR #136, opened 2022-05-31
prerequisite setup for ubuntu

ritvik-verma:
since only macOS setup was available.


myungjin (approved):
@ritvik-verma Thanks for the PR.

PR #141, opened 2022-06-07
feature: log artifact to mlflow

GaoxiangLuo:
This is a simple yet effective feature from mlflow that allows user to add artifact of any types of data such as png, csv, pickle, etc. Recommend to be included in the next release.

type | file | changes
-- | -- | --
modified | lib/python/flame/registry/mlflow.py | add api for logging artifacts to MLFLow


myungjin (approved):
lgtm

PR #144, opened 2022-06-10
exps: benchmark adaptive server optimizers

GaoxiangLuo:
In this commit, we create an example of using different adaptive server optimizers to show the benefit of not staying with FedAvg algorithm, and also Flame system provides the flexibility of designing your own server optimizer and using it on the fly.

type | file | changes
-- | -- | --
modified | examples/medmnist/README.md | docs
deleted | examples/medmnist/dataset.json | 
created | examples/medmnist/dataset1.json | dataset example
created | examples/medmnist/dataset2.json | dataset example
created | examples/medmnist/dataset3.json | dataset example
created | examples/medmnist/dataset4.json | dataset example
created | examples/medmnist/dataset5.json | dataset example
created | examples/medmnist/dataset6.json | dataset example
created | examples/medmnist/dataset7.json | dataset example
created | examples/medmnist/dataset8.json | dataset example
created | examples/medmnist/dataset9.json | dataset example
created | examples/medmnist/dataset10.json | dataset example
created | examples/medmnist/images/train_summary.png | image
created | examples/medmnist/images/val_summary.png | image
modified | examples/medmnist/job.json | job json
modified | examples/medmnist/medmnist.zip | code
modified | examples/medmnist/schema.json | schema


examples/medmnist/dataset1.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "MedMNIST dataset1",
+
+    "description": "Dataset Repo: https://github.com/MedMNIST/MedMNIST, NonIID generating algorithm from FedMA paper repo",
+
+    "url": "https://www.googleapis.com/drive/v3/files/14SVYg6irtBtwfoeYBT24VhS2-n87men9?alt=media&key=${Google Drive API Key}",
```
myungjin, line 6: I think `${Google Drive API Key}` can be specified an environment variable (e.g., GDRIVE_API_KEY=some_api_key). And then, a user can do something like the following: `export $GDRIVE_API_KEY`.

In this way, user doesn't need to modify all 10 files.

examples/medmnist/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Step 3: add code to the design
+
+```bash
+flamectl create code medmnist.zip --design medmnist
+```
+
+### Step 4: create datasets
+
+Thanks to NVFlare providing [the NonIID dataset generation script](https://github.com/NVIDIA/NVFlare/blob/main/examples/cifar10/pt/utils/prepare_data.py) proposed by [FedMA](https://github.com/IBM/FedMA). We used the script to split the PathMNIST dataset of MedMNIST into 10 non-overlapping portions in a Non-IID fashion. And for each individual dataset, we splitted it into training and validation set in a 8:2 ratio. The following is the data distribution of the training set of all clients:
```
myungjin, line 29: Can you simply say something like the following? `we use NVFlare's NonIID dataset generation script ...` or `there is NonIID dataset generation script from NVFlare`?

examples/medmnist/README.md
```diff
@@ -0,0 +1,116 @@
...
+![train_summary](images/train_summary.png)
+And the following is the data distribution of the validation set of all clients:
+![val_summary](images/val_summary.png)
+
+As the heterogenous split of PathMNIST datasets are stored at Google Drive, it requires the user to generate a Google Drive API key and replace it with the `${Google Drive API Key}` in the URL.
```
myungjin, line 34: It may be useful to add a link or text to describe how to create an api key.

Also, setting an environment variable (e.g., GDRIVE_API_KEY) would be more convenient than asking to edit all dataset files.

examples/medmnist/README.md
```diff
@@ -0,0 +1,116 @@
+## MedMNIST
+
+We use the PathMNIST dataset from MedMNIST to go over an example of using adaptive aggregator on a data heterogeneity setting, implemented in PyTorch. This example is run within [fiab](../../docs/03-fiab.md) environment.
```
myungjin, line 3: Perhaps we need to cite their paper since the dataset is licensed under Creative Commons Attribution 4.0 International ([CC BY 4.0](https://creativecommons.org/licenses/by/4.0/)).

myungjin (commented):
left a few comments.

examples/medmnist/dataset1.json
GaoxiangLuo: I eventually used a public GitHub repo to host the data.

examples/medmnist/dataset1.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "MedMNIST dataset1",
+
+    "description": "Dataset Repo: https://github.com/MedMNIST/MedMNIST, NonIID generating algorithm from FedMA paper repo",
+
+    "url": "https://github.com/GaoxiangLuo/flame-datasets/blob/main/site1.npz",
```
myungjin, line 6: Did you test this url? With curl, it won't be download. The url gives an html file.

A working url should have a form of the following: https://raw.githubusercontent.com/GaoxiangLuo/flame-datasets/main/site1.npz

myungjin (commented):
One comment left.

examples/medmnist/dataset1.json
GaoxiangLuo: Maybe that's why it doesn't work. I tested it with `wget` and it worked. And thanks for the working url!

GaoxiangLuo: Actually, `https://raw.githubusercontent.com/GaoxiangLuo/flame-datasets/main/site1.npz` won't work either, but `https://raw.github.com/GaoxiangLuo/flame-datasets/main/site1.npz` will work.  If not add the `raw.` and remove the `/blob`, it's getting the full GitHub view of that file, not just the contents.

myungjin (approved):
lgtm

PR #147, opened 2022-06-16
[control-plane] First commit for deployer as separate pod

dhruvsgarg:
This initiates an important change for the flame controlplane. We will be developing a separate `deployer` service to de-couple the current `deployer` from `controller`. A separate `deployer` service required to add support for Flame to run across multiple clusters.

This first commit creates a separate `pod` for deployer that is perpetually in a `running` state. To enable this, a placeholder app for the deployer was created and the helm-chart files were also added. The app will be implemented in the future commits to expose all the required services for the deployer.
Deployer_pod

### DIFF
--- a/cmd/deployer/app/appserver.go
+++ b/cmd/deployer/app/appserver.go
@@ -1,0 +1,1 @@
+package app
--- a/cmd/deployer/app/service.go
+++ b/cmd/deployer/app/service.go
@@ -1,0 +1,1 @@
+package app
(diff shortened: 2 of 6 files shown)


GaoxiangLuo (commented):
lgtm. It creates a deployer pod as a placeholder, and the pod is created with `./flame.sh start` and shutdown with `./flame.sh stop`. I run the MNIST example just fine, and it behaves as usual. In the future, the deployer pod may decouple the functions of controller.

myungjin (commented):
left several comments.

myungjin (approved):
lgtm

PR #148, opened 2022-06-17
feature: fetch latest weights in distributed learning

GaoxiangLuo:
This commit allows a new trainer to fetch the latest global weight of distributed learning. The weight can either be 1) from the global weights produced by the ring all-reduce algorithm or 2) from a single trainer if there is only one trainer previsouly. Either case the trainer that holds the current best weights will be the committer.

type | file | changes
-- | -- | --
modified | lib/python/flame/mode/distributed/trainer.py | use existing communication channel to fetch latest weights
modified | lib/python/flame/mode/message.py |  add a new message type

### DIFF
--- a/lib/python/flame/mode/distributed/trainer.py
+++ b/lib/python/flame/mode/distributed/trainer.py
@@ -33,11 +33,6 @@ logger = logging.getLogger(__name__)
 
 TAG_RING_ALLREDUCE = 'ring_allreduce'
 
-# the number of copies to save parameters and model artifact
-# this is for redundancy
-COMMIT_COUNT = 3
-
-
 class Trainer(Role, metaclass=ABCMeta):
     """Trainer implements an ML training role."""
 
@@ -63,6 +58,7 @@ class Trainer(Role, metaclass=ABCMeta):
         if base_model and base_model.name != "" and base_model.version > 0:
             self.model = self.registry_client.load_model(
                 base_model.name, base_model.version)
+        self.ring_weights = None # store the latest model weights from ring all-reduce
 
         self.registry_client.setup_run(mlflow_runname(self.config))
         self.metrics = dict()
@@ -71,7 +67,7 @@ class Trainer(Role, metaclass=ABCMeta):
         self._rounds = self.config.hyperparameters['rounds']
         self._work_done = False
 
-        self.is_commiter = False
+        self.is_committer = False
 
         self.framework = get_ml_framework_in_use()
         if self.framework == MLFramework.UNKNOWN:
@@ -105,7 +101,6 @@ class Trainer(Role, metaclass=ABCMeta):
         logger.debug(f"total_data_count: {total_data_count}")
         if not success:
             # members don't agree, we can't do ring-allreduce
-            self.is_commiter = True
             return
 
         self._update_weights()
@@ -164,8 +159,8 @@ class Trainer(Role, metaclass=ABCMeta):
 
             recv_chunk_idx = (rank - i - 1 + size) % size
             msg = channel.recv(ends[recv_from])
-            if MessageType.WEIGHTS not in msg:
-                logger.error(f"end_id: {ends[recv_from]}, msg: {msg}")
+            while MessageType.WEIGHTS not in msg:
+                msg = channel.recv(ends[recv_from])
             recv_chunk = msg[MessageType.WEIGHTS]
 
             logger.debug(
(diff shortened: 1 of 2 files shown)


lib/python/flame/mode/distributed/trainer.py
```diff
@@ -320,13 +326,20 @@ def _member_check(self, channel) -> tuple[bool, int]:
...
             return False, 0
 
         msg = {
             MessageType.MEMBER_DIGEST: digest,
             MessageType.DATASET_SIZE: self.dataset_size,
             MessageType.ROUND: self._round
         }
+        # only commiter will send the latest global weights instead of all sending, it saves bandwidth
+        if self.is_commiter:
+            msg[MessageType.RING_WEIGHTS] = self.ring_weights
```
myungjin, line 342: This means that weights are broadcast every round. This eliminates the benefit of distributed learning.

For example, if there is no new worker joining late, there is no need for broadcasting the weights because all workers have the latest model weights computed from the previous round. But this still broadcasts the model weights.

Also, broadcast is also expensive because only new workers need the latest model weights while other existing workers don't.

myungjin (commented):
The proposed approach is quite expensive. An efficient approach is needed.

lib/python/flame/mode/distributed/trainer.py
GaoxiangLuo: Thank you for your comments. It's indeed very expensive to broadcast the weights every round. The current implementation is changed to the follows: a new trainer will broadcast a message indicating the arrival of a itself, and then whoever other trainer receives the broadcast message and meanwhile is the committer, will send the latest optimal weights to the new trainer. If there isn't any committers, which happens when more than one trainer gets initiated at the same time, they will proceed to ring all-reduce operation.

lib/python/flame/mode/distributed/trainer.py
```diff
@@ -33,11 +33,6 @@
 
 TAG_RING_ALLREDUCE = 'ring_allreduce'
 
-# the number of copies to save parameters and model artifact
-# this is for redundancy
-COMMIT_COUNT = 3
```
GaoxiangLuo, line 38: I removed this global variable because the current design only works when there is at most one committer at a time.

lib/python/flame/mode/distributed/trainer.py
```diff
@@ -71,7 +67,7 @@ def internal_init(self) -> None:
         self._rounds = self.config.hyperparameters['rounds']
         self._work_done = False
 
-        self.is_commiter = False
+        self.is_committer = False
```
GaoxiangLuo, line 70: refactor; same as below.

lib/python/flame/mode/distributed/trainer.py
```diff
@@ -164,8 +159,8 @@ def _do_ring_allreduce(self, channel):
 
             recv_chunk_idx = (rank - i - 1 + size) % size
             msg = channel.recv(ends[recv_from])
-            if MessageType.WEIGHTS not in msg:
-                logger.error(f"end_id: {ends[recv_from]}, msg: {msg}")
+            while MessageType.WEIGHTS not in msg:
```
GaoxiangLuo, line 162: This change aims to improve synchronization. Because a digest message of a new trainer can be sent out when the others are already in the ring-reduce phase.

lib/python/flame/mode/distributed/trainer.py
```diff
@@ -309,6 +309,16 @@ def _handle_member_check(self, channel, end, digest) -> tuple[bool, int]:
 
             dataset_size = msg[MessageType.DATASET_SIZE]
 
+            # new trainer fetches the latest weights from the committer
+            if msg[MessageType.IS_COMMITTER]:
```
GaoxiangLuo, line 313: If there aren't any committer, which happens several trainers start at the same time, it will proceed ring all-reduce as usual.

lib/python/flame/mode/distributed/trainer.py
myungjin: I am hoping to handle this part without while loop and recv function as this part can cause some additional complexity and issue in the future.

While we keep this as is, we can modify the other place.

We can leverage _member_check(). Near the end of that function (e.g., https://github.com/cisco-open/flame/pull/148/files#diff-52c4d612c73519f6889ec671bb33ba9d6002f30f01bf76e6132e069d5115c3eaR371), we can save `ends` in an instance variable, e.g. `self.ends = ends`.

Then, when we call _do_ring_allreduce(), we no longer need https://github.com/cisco-open/flame/pull/148/files#diff-52c4d612c73519f6889ec671bb33ba9d6002f30f01bf76e6132e069d5115c3eaR119-R121 lines.

Within the _do_ring_allreduce(), we replace `ends` with `self.ends`. That way, we don't need to worry about new workers sending non-weights message.

The self.ends are the workers who agreed on a ring topology. So, we can save them and reuse them instead of calling channel.ends() in _do_ring_allreduce().

I hope this makes sense. Otherwise, let's discuss offline.

myungjin (commented):
left one comment

myungjin (approved):
approving this PR based on our discussion. please create a new pr to incorporate the `self.ends` idea we discussed.

PR #156, opened 2022-06-24
Job template as configmap

myungjin:
To run a training job, controller relies on a helm chart under
templates/job. Some variables (e.g., endpoints of apiserver and
notifier) of files in the folder are hard coded. When those endpoints
need to be updated, a new image needs to be built, which is time
consuming. Also, a user needs to modify several files, which can cause
misconfigurations.

To address the problem, now the templates are moved into configmaps so
that it can be updated easily at deployment time.


GaoxiangLuo (approved):
lgtm. It uses placeholder instead of hard code to make various setting including end point adaptive, yielding the ease of deployment in the future. And it doesn't change the current behavior of the library (i.e., demo example still runs).

dhruvsgarg (approved):
lgtm, tested.

PR #157, opened 2022-06-24
Prepending API endpoints involving user to prevent mismatches with future endpoints

dhruvsgarg:
We will be adding new API endpoints for multi-cluster support. These new API endpoints will follow the pattern `/computes/{..}`. However many existing API endpoints follow the pattern `/{user}/{..}`. Thus, there is a possible issue of incorrect api resolution when a `{user}` name is `computes`. To clean up the endpoints and to avoid such cases, endpoints containing `/{users}/` are prepended with a static `/users/`. This will ensure correct end-point resolution.

### DIFF
--- a/api/dataset_api.partials.yml
+++ b/api/dataset_api.partials.yml
@@ -56,7 +56,7 @@
     tags:
       - datasets
 
-/{user}/datasets:
+/users/{user}/datasets:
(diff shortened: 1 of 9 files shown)


myungjin (approved):
lgtm

PR #160, opened 2022-06-28
misc: minor update of api spec

myungjin:
In job components partials, the types of some properties are incorrect,
which caused  warning messages when openapi-generator is in use. These
are handled in this change.


pkg/openapi/model_base_model.go
```diff
@@ -30,3 +30,30 @@ type BaseModel struct {
...
+	elements := map[string]interface{}{
+		"name":    obj.Name,
+		"version": obj.Version,
+	}
+	for name, el := range elements {
+		if isZero := IsZeroValue(el); isZero {
+			return &RequiredError{Field: name}
+		}
+	}
+
```
dhruvsgarg, line 45: This part doesn't seem to be handled in the current code or the specification of the base model. The creation of job using `/examples/mnist/` is following. 
image

myungjin: Thanks for catching this. I will make those variables optional.

dhruvsgarg: Thanks

dhruvsgarg (approved):
lgtm

PR #161, opened 2022-06-29
optimize: reduce distributed learning code complexity

GaoxiangLuo:
This commit reduces the complexity of the distributed learning code by storing the ends who agree with the ring topology at a time, so that a trainer in a ring won't send weights message to trainers outside of the ring. Meanwhile, a new trainer's arrival message won't send to the trainers in a ring either. Hence, the moment that a new trainer fetches the global optimal weights will be when the new trainer is in the ring.

type | file | changes
-- | -- | --
modified | lib/python/flame/mode/distributed/trainer.py | store ends in a ring separately


myungjin (approved):
Thanks! Looks great to me!

PR #163, opened 2022-06-30
feature: minio support in fiab

myungjin:
In fiab, mlflow is configured without an object store. Therefore,
artifacts (e.g., model) are stored locally within a container. Once
training is over, then the artifacts get lost. With support for
minio (an open-source object store), now artifacts are saved in an
object store. Also, minio's endpoint is exposed via ingress so that
workers outside a cluster can save artifacts. Note that tls for minio
is not enabled because selfsigned cert used in fiab is likely to make
mlflow's client complain about a verification issue.

A caveat is that no persistent volume is configured in this minio
setup, artifacts will get lost if minio pod crashes or terminates. The
support for minio is for reference so that users can configure minio
or aws s3 correctly for production.

### DIFF
--- a/fiab/helm-chart/Chart.yaml
+++ b/fiab/helm-chart/Chart.yaml
@@ -25,3 +25,7 @@ dependencies:
   - name: mongodb
     version: 10.31.3
     repository: https://charts.bitnami.com/bitnami
+
+  - name: minio
+    version: 4.0.2
+    repository: https://charts.min.io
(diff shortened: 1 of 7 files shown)


GaoxiangLuo (commented):
I can see a new pod named minio initialized. While running with MNIST example, the artifact full path starts with `s3://mlruns/XX`, which is as expected. If I login to the minio pod, where are the artifacts stored? Also, are we supposed to be able to open `http://minio.flame.test`?

myungjin:
@GaoxiangLuo you should be able to download artifacts from mlflow UI.

GaoxiangLuo:
>

GaoxiangLuo (approved):
lgtm

PR #166, opened 2022-07-05
typo: use correct func tags for distributed learning

GaoxiangLuo:
Previously, the distributed learning topology works with python library but doesn't work with flame system. It turns out to be a simple typo error. Now fixed.

type | file | changes
-- | -- | --
modified | examples/distributed_training/schema.json | update func tags in schema


myungjin (approved):
lgtm

PR #167, opened 2022-07-06
fix: message mixed up in distributed learning mode

myungjin:
During the member check, there is a high chance to receive weights
message because other workers started ring all-reduce process while
a trainer is in a member check loop. In such a case, the current code
will delcare failure of the membership check. This creates an
inconsistent agreement among workers, meaning that some think they are
in a ring and some others don't. To address the issue, we reinstate
the peek statement.

In addition, some cosmetric reformatting is done.

### DIFF
--- a/lib/python/flame/examples/dist_mnist/trainer/config1.json
+++ b/lib/python/flame/examples/dist_mnist/trainer/config1.json
@@ -45,7 +45,7 @@
     },
     "registry": {
 	"sort": "mlflow",
-	"uri": "http://flame-mlflow:5000"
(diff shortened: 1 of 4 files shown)


dhruvsgarg (approved):
lgtm

GaoxiangLuo:
@myungjin Right. After I changed the config, and ran the example with flame system, I didn't see the issue. "dummy" sounds like a good placeholder for the registry to me.

myungjin:
@GaoxiangLuo OK. Task for `dummy` registry should a separate PR. So, I will merge this one as is.

PR #170, opened 2022-07-07
feature: hybrid mode with ring allreduce

myungjin:
Hybrid mode of combining distributed learning and federated learning
is implemented. This implementation is based on PR #131.

The current implementation has a bug thta can cause a deadlock when
trainers arrive late. Addressing the issue will be handled as a
separate PR as it requires quite changes in backend, channel, channel
manager, etc.

### DIFF
--- a/lib/python/flame/mode/hybrid/trainer.py
+++ b/lib/python/flame/mode/hybrid/trainer.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 #
 # SPDX-License-Identifier: Apache-2.0
-
 """hybrid FL trainer."""
 
 import logging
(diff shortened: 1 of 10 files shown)


lib/python/flame/mode/hybrid/trainer.py
```diff
@@ -28,69 +27,85 @@
...
+            logger.debug("ring was not formed; each shares its weights")
+            weights, size = self.weights, self.dataset_size
+
+        # reaching here means that all-reduce took place. So, a committer
+        # only needs to share model weights (built from all-reduce) and
+        # the total number of data samples used in a ring.
+        #
+        # non-committers send a dummy message so that the aggregator won't
+        # be blocked.
+        # TODO: figure out a way not to send a dummy message
```
GaoxiangLuo, line 98: Isn't line 105 already a way to send the dummy message?

GaoxiangLuo (approved):
lgtm. The implementation is consistent with the previous PR. 

When I was testing it:
1. If a trainer of `us/org` is already running with the aggregator, then when a new trainer of `us/XXX` joins, it will cause a deadlock. 
2. However, when a new trainer of `eu/XXX` joins, it won't cause a deadlock if there are not existing `eu/XXX` trainers. 

In a word, a new trainer will only be hanging if there is an existing trainer of the same realm who may potentially form a ring with the new trainer. The new trainer is blocked on the stage of fetching weights from the aggregator. Hence, it may have something to do with how to distinguish aggregator-trainer channel and trainer-trainer channel. This will be resolved as a separate PR.

PR #172, opened 2022-07-11
fix: deadlock in hybrid fl mode

myungjin:
A deadlock situation often occur when trainers in the same group
arrive later. This is because there is a circular dependancy on
message recv calls across trainers and aggregator. More specifically,
trainers which already received a global model from an aggregator
wait for membership check messages, but trainers which didn't receive
a global model will keep waiting to receive a global model without
sending membership check message. This creates a deadlock.

We break the circular dependancy by making trainers join a channel for
distributed learning only when they receive a global model from an
aggregator.


GaoxiangLuo (approved):
The committer works fine and the reason of not saving model to MLFlow is that hybrid example's `config.json` is using dummy registry instead of MLFlow registry. Besides, it looks to me.

PR #184, opened 2022-07-14
harden flame.sh stop process

myungjin:
if "flame.sh stop" is executed more than once, the current script
destroys configuration of coredns. As a resolution, we delete last 5
lines of the config if a flame.test section exists in coredns config.


dhruvsgarg (approved):
lgtm. Tested multiple flame-start and flame-stop. The script no longer deletes extra lines from core-dns on multiple consecutive flame-stop calls.

PR #188, opened 2022-07-18
Redefine proto for job notification

dhruvsgarg:
In the future, proto will also include deployer notifications / events. Hence this PR updates current set of events to specify that they are job events.


myungjin (approved):
lgtm

PR #194, opened 2022-07-21
Adding few restAPIs for deployer

dhruvsgarg:
This pr adds restAPIs for the deployer to be able to exchange information with the controller.


api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
               $ref: '#/components/schemas/ComputeSpec'
         description: Information about compute cluster
+      default:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
+        description: unexpected error
+
+/computes/{computeId}/deployments:
```
myungjin, line 172: which component is going to use this resource? I couldn't immediately think about which component is going to consume it.

If there is no component, let's not add it.

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+      default:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
+        description: unexpected error
+
+/computes/{computeId}/deployments/{jobId}/config:
+  ########################################
+  # Get config for a job deployment on a compute cluster
```
myungjin, line 208:
```suggestion
  # Get deployment config for a job on a compute cluster
```

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+  ########################################
+  get:
+    operationId: getComputeJobConfig
```
myungjin, line 211: How about the following?
```suggestion
    operationId: getDeploymentConfig
```

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+    summary: Get config for a job deployment on a compute cluster
```
myungjin, line 212:
```suggestion
    summary: Get deployment config for a job on a compute cluster
```

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+        explode: false
+        required: true
+    responses:
+      "200":
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                $ref: '#/components/schemas/AgentConfig'
```
myungjin, line 239: Let's make names consistent between resource name and response message. So, instead of defining an array here, why don't you create a component type called `DeploymentConfig`?

In the DeploymentConfig, you can define the following:
jobId, imageLoc and a list of (taskId, taskKey) pairs.

This way, we can avoid the repetition of jobId and imageLoc.
And (taskId, taskKey) can be defined as a separate component if you like.

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+        description: A paged array of configurations for agents within a deployment for a job id
```
myungjin, line 240: you don't need a paged array. we need to get all of the necessary configuration details. Please revise this.

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+      default:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
+        description: unexpected error
+
+/computes/{computeId}/deployments/{jobId}/status:
+  ########################################
+  # Get status of a job deployment on a compute cluster
```
myungjin, line 250:
```suggestion
  # Get deployment status of a job on a compute cluster
```

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+  ########################################
+  get:
+    operationId: getComputeJobStatus
```
myungjin, line 253:
```suggestion
    operationId: getDeploymentStatus
```

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+    summary: Get the status for a job deployment on a compute cluster
```
myungjin, line 254:
```suggestion
    summary: Get the deployment status for a job on a compute cluster
```

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
+        description: unexpected error
+
+/computes/{computeId}/deployments/{jobId}/status:
+  ########################################
+  # Get status of a job deployment on a compute cluster
+  ########################################
+  get:
```
myungjin, line 252: who is going to update deployment status? `get` api is not useful if there is no `post` in this case.

Also, try to answer if this resource is needed by considering which component is going to consume the resource.

Note that I am suggesting that this resource is unnecessary.

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+        required: true
+    responses:
+      "200":
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                $ref: '#/components/schemas/AgentStatus'
+        description: A paged array of status for agents within a job deployment
```
myungjin, line 282: To enable a paged array, an additional variable (i.e., length) needs to be defined.  But, it may not be needed at the moment. Please consider to revise the description.

api/computes_api.partials.yml
```diff
@@ -162,6 +162,124 @@
...
+        explode: false
+        required: true
+    responses:
+      "200":
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                $ref: '#/components/schemas/AgentStatus'
```
myungjin, line 281: Similar to the previous comment on AgentConfig, please consider to make naming consistent.

So, avoid using array and define DeploymentStatus.
DeploymentStatus can have the following elements:
- jobId
- deployed (i.e., an array of deployed agents)
- revoked (i.e., an array of revoked agents)
- failed (i.e., an array of failed agents).

agent corresponds to taskId.

myungjin: Also, this part (an array of AgentStatus) is almost exactly same as `ComputeJobStatus`. So, we need to avoid redundant definition of a similar concept.
If you decide to use `ComputeJobStatus`, then it needs to be renamed as `DeploymentStatus`.

myungjin (commented):
thanks for the PR. defining api spec well is a hard task. I left some comments. please consider to accommodate them.

api/computes_api.partials.yml
dhruvsgarg: I created this so that the admin (using flamectl) can get the list of all running jobs within a compute cluster

api/computes_components.partials.yml
```diff
@@ -63,4 +63,68 @@ ComputeState:
...
+    $ref: '#/components/schemas/AgentState'
+
+########################################
+# Agent state
+########################################
+AgentState:
+  enum:
+    - deployed
+    - revoked
+    - error
```
myungjin, line 130: error is also too generic; and also `error` is a builtin object in golang.
Perhaps a better name is to add a prefix (e.g., `agent`)

- agent_deployed
- agent_revoked
- agent_failed

Note that this idea of adding a prefix applies to all the enums defined in the api specification.

myungjin (approved):
I am approving this PR while there is one more minor comment. Please feel free to visit this comment in a future PR.

PR #196, opened 2022-07-22
Adding apikey to deployer apis

dhruvsgarg:
This pr adds apiKey parameter to deployer restAPIs. It is the first step towards authenticating the deployers while they add or revoke resources for deployments.

### DIFF
--- a/api/computes_api.partials.yml
+++ b/api/computes_api.partials.yml
@@ -58,6 +58,13 @@
         style: simple
         explode: false
         required: true
+      - in: header
+        name: X-API-KEY
+        schema:
+          type: string
+        style: simple
+        explode: false
+        required: true
     responses:
       "200":
         content:
@@ -84,6 +91,13 @@
       - name: computeId
         description: compute id of compute cluster
         in: path
+        schema:
+          type: string
+        style: simple
+        explode: false
+        required: true
+      - in: header
+        name: X-API-KEY
         schema:
           type: string
         style: simple
@@ -123,6 +137,13 @@
         style: simple
         explode: false
         required: true
+      - in: header
+        name: X-API-KEY
+        schema:
+          type: string
+        style: simple
+        explode: false
+        required: true
     responses:
       '200':
         description: Deleted
@@ -155,6 +176,13 @@
         style: simple
         explode: false
         required: true
+      - in: header
+        name: X-API-KEY
+        schema:
+          type: string
+        style: simple
+        explode: false
+        required: true
     responses:
       "200":
         content:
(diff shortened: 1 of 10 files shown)


api/computes_api.partials.yml
```diff
@@ -40,7 +40,7 @@
               $ref: '#/components/schemas/Error'
         description: unexpected error
 
-/computes/{computeId}:
+/computes/{computeId}/apikey/{apiKey}:
```
myungjin, line 43: don't create apikey as a resource. apikey should be included in one of three ways: query, header or cookie. Out of three ways, I suggest to use header.

Refer to https://swagger.io/docs/specification/authentication/api-keys/ to define apikey correctly in the openapi specification.

myungjin (changes requested):
left one comment. please address it.

api/computes_api.partials.yml
```diff
@@ -58,6 +58,8 @@
         style: simple
         explode: false
         required: true
+    security:
+      - ApiKeyAuth: []
```
myungjin, line 62: Adding securitySchemes and security don't seem to work. In the generated code, no code for parsing api-key is generated.

Let's try to apply a workaround.

Use the following under `parameters` section of methods of interest.
```
      - in: header
        name: X-API-KEY
        schema:
          type: string
```

Use X-API-KEY as is; it is a common key name.

A generated code would have something like:
```
	xAPIKEYParam := r.Header.Get("X-API-KEY")
	result, err := c.service.GetDeploymentStatus(r.Context(), computeIdParam, jobIdParam, xAPIKEYParam)
```

myungjin (commented):
openapi-generator doesn't seem to handle securitySchemes and security properly. For now, as a workaround, we probably need to apply a workaround. 

Please check out detailed comments.

api/computes_securitySchemes.partials.yml
```diff
@@ -0,0 +1,24 @@
+# Copyright 2022 Cisco Systems, Inc. and its affiliates
```
myungjin, line 1: If you want to keep this file, the filename still has a camelCase style. Please rename it with lowercases only. This will cause an error in linux since you used a lowercase version of the file name in  openapi.yaml.mustache.

api/computes_securitySchemes.partials.yml
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+########################################
+# Security scheme
+########################################
+ApiKeyAuth:
+  type: apiKey
+  description: API key to authorize requests
+  in: header
+  name: ApiKey
```
myungjin, line 24:
```suggestion
  name: X-API-KEY
```

myungjin (commented):
two comments are left.

myungjin (approved):
lgtm

PR #200, opened 2022-07-29
Added checks for dataset registration

dhruvsgarg:
This pr maps the realm in the input dataset json to the registered computes. The realm in the dataset json specifies either region OR computeId. If the dataset is private, the specified region or computeId must be correct. If it is invalid, an error would be thrown and creation of the dataset would fail. In the case of public datasets, if the region or computeId does not match, the defaultRealm is selected which also specifies a default computeId. The dataset realm is used at the time of task creation to assign computeIds and the job tasks get assigned to that compute cluster.

### DIFF
--- a/cmd/controller/app/database/mongodb/compute.go
+++ b/cmd/controller/app/database/mongodb/compute.go
@@ -26,19 +26,26 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/cisco-open/flame/pkg/openapi"
+	"github.com/cisco-open/flame/pkg/util"
 )
 
 // RegisterCompute creates a new cluster compute specification and returns ComputeStatus
 func (db *MongoService) RegisterCompute(computeSpec openapi.ComputeSpec) (openapi.ComputeStatus, error) {
 	// First check if the compute was previously registered
-	filter := bson.M{"computeid": computeSpec.ComputeId}
+	filter := bson.M{util.DBFieldComputeId: computeSpec.ComputeId}
 	checkResult := db.computeCollection.FindOne(context.TODO(), filter)
+	if (checkResult.Err() != nil) && (checkResult.Err() != mongo.ErrNoDocuments) {
+		errMsg := fmt.Sprintf("Failed to register compute : %v", checkResult.Err())
+		zap.S().Errorf(errMsg)
+
+		return openapi.ComputeStatus{}, ErrorCheck(checkResult.Err())
+	}
 	if checkResult.Err() == mongo.ErrNoDocuments {
 		// If it was not registered previously, need to register
 		result, err := db.computeCollection.InsertOne(context.TODO(), computeSpec)
 		if err != nil {
-			zap.S().Errorf("Failed to register new compute in database: result: %v, error: %v", result, err)
-
+			errMsg := fmt.Sprintf("Failed to register new compute in database: result: %v, error: %v", result, err)
+			zap.S().Errorf(errMsg)
 			return openapi.ComputeStatus{}, ErrorCheck(err)
 		}
 
@@ -100,7 +107,7 @@ func (db *MongoService) UpdateComputeStatus(computeId string, computeStatus open
 		setElements[dateKey] = updateTime
 	}
 
-	filter := bson.M{"computeid": computeId}
+	filter := bson.M{util.DBFieldComputeId: computeId}
 	update := bson.M{"$set": setElements}
 
 	updatedDoc := openapi.ComputeStatus{}
@@ -110,4 +117,59 @@ func (db *MongoService) UpdateComputeStatus(computeId string, computeStatus open
 	}
 
 	return updateTime, nil
+}
+
+func (db *MongoService) GetComputeIdsByRegion(region string) ([]string, error) {
+	zap.S().Infof("get all computes in the region: %s", region)
+
+	filter := bson.M{util.DBFieldComputeRegion: region}
+	cursor, err := db.computeCollection.Find(context.TODO(), filter)
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to fetch computes in the region: %s, err : %v", region, err)
+		zap.S().Errorf(errMsg)
+
+		return []string{}, fmt.Errorf(errMsg)
+	}
+
+	defer cursor.Close(context.TODO())
+	var computeIdList []string
+
+	for cursor.Next(context.TODO()) {
+		var computeSpec openapi.ComputeSpec
+		if err = cursor.Decode(&computeSpec); err != nil {
+			errMsg := fmt.Sprintf("failed to decode compute spec with error: %v", err)
+			zap.S().Errorf(errMsg)
+
+			return []string{}, ErrorCheck(err)
+		}
+
+		computeIdList = append(computeIdList, computeSpec.ComputeId)
+	}
+
+	if len(computeIdList) == 0 {
+		errMsg := fmt.Sprintf("could not find any computes for the region: %s", region)
+		zap.S().Errorf(errMsg)
+
+		return []string{}, fmt.Errorf(errMsg)
+	}
+	return computeIdList, nil
+}
+
+func (db *MongoService) GetComputeById(computeId string) (openapi.ComputeSpec, error) {
+	filter := bson.M{util.DBFieldComputeId: computeId}
+	checkResult := db.computeCollection.FindOne(context.TODO(), filter)
+	if checkResult.Err() != nil {
+		errMsg := fmt.Sprintf("failed to find a compute with computeId: %s", computeId)
+		zap.S().Errorf(errMsg)
+		return openapi.ComputeSpec{}, fmt.Errorf(errMsg)
+	}
+
+	var currentDocument openapi.ComputeSpec
+	err := checkResult.Decode(&currentDocument)
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to parse currentDocument: %v", err)
+		zap.S().Errorf(errMsg)
+		return openapi.ComputeSpec{}, fmt.Errorf(errMsg)
+	}
+	return currentDocument, nil
 }
--- a/pkg/openapi/controller/api_datasets_service.go
+++ b/pkg/openapi/controller/api_datasets_service.go
@@ -30,9 +30,14 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"path"
+	"strings"
+
+	"go.uber.org/zap"
 
 	"github.com/cisco-open/flame/cmd/controller/app/database"
 	"github.com/cisco-open/flame/pkg/openapi"
+	"github.com/cisco-open/flame/pkg/util"
 )
 
 // DatasetsApiService is a service that implents the logic for the DatasetsApiServicer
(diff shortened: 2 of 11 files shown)


cmd/controller/app/database/mongodb/compute.go
```diff
@@ -111,3 +111,56 @@ func (db *MongoService) UpdateComputeStatus(computeId string, computeStatus open
 
 	return updateTime, nil
 }
+
+func (db *MongoService) CheckComputeByRegion(region string) ([]string, error) {
```
myungjin, line 115: The function name is misleading. How about the following?
```suggestion
func (db *MongoService) GetComputesByRegion(region string) ([]string, error) {
```

cmd/controller/app/database/mongodb/compute.go
```diff
@@ -111,3 +111,56 @@ func (db *MongoService) UpdateComputeStatus(computeId string, computeStatus open
...
+	checkResult := db.computeCollection.FindOne(context.TODO(), filter)
+	if checkResult.Err() == mongo.ErrNoDocuments {
+		// No computes found for in this region
+		errMsg := fmt.Sprintf("Failed to find a compute in region: %s", region)
+		zap.S().Errorf(errMsg)
+		return []string{}, fmt.Errorf(errMsg)
+	}
+
+	// If computes are found in the region, we can return a list of computeIds. Sending one computeId for now
+	// TODO send more computeIds instead of just one. Let the calling code decide which computeId to use
```
myungjin, line 126: Parsing more than one doc shouldn't be too hard.
Please consider to finish the implementation instead of leaving TODO.

Some example code can be found https://github.com/cisco-open/flame/blob/main/cmd/controller/app/database/mongodb/job.go#L123

cmd/controller/app/database/mongodb/compute.go
```diff
@@ -111,3 +111,56 @@ func (db *MongoService) UpdateComputeStatus(computeId string, computeStatus open
...
+	if checkResult.Err() == mongo.ErrNoDocuments {
+		// No computes found with the same computeId
+		errMsg := fmt.Sprintf("Failed to find a compute with computeId: %s", computeId)
+		zap.S().Errorf(errMsg)
+		return false, fmt.Errorf(errMsg)
+	}
+	return true, nil
+}
+
+func (db *MongoService) GetComputeRegion(computeId string) (string, error) {
```
myungjin, line 150: Instead of returning `region` string only, perhaps returning `ComputeSpec` is more useful for future use.

The caller of this function can determine which attributes in ComputeSpect to use. Here we can keep DB API relatively simple.

If you decide to return the whole ComputeSpec, please rename the function (e.g., `GetCompute` or `GetComputeById`)

cmd/controller/app/job/default_handler.go
```diff
@@ -323,8 +324,28 @@ func (h *DefaultHandler) notifyJob(evtType pbNotify.JobEventType) error {
 }
 
 func (h *DefaultHandler) notifyDeploy(evtType pbNotify.DeployEventType) error {
-	// TODO: add computeId field to tasks and remove hard-coding
-	computeIds := []string{"compute-1"}
+	// TODO implement for FromUser datasets
```
myungjin, line 327: `FromUser` doesn't need notification because bringing up compute resource is responsibility of user.

cmd/controller/app/job/default_handler.go
```diff
@@ -323,8 +324,28 @@ func (h *DefaultHandler) notifyJob(evtType pbNotify.JobEventType) error {
...
+	computeIdMap := make(map[string]bool)
+
+	// Need to get a list of computeIds to notify for deployment
+	// To do this, get all datasets for the jobId, iterate and get computeId from datasetRealm
+	fromSystemDatasets := h.jobSpec.DataSpec.FromSystem
```
myungjin, line 332: notification shouldn't be based on datasets. It should be based on tasks in the task collection given a job ID.

I think we talked about this before. I thought that when task is created in DB, task will have a compute ID in its DB record.
So, when tasks are retrieved, compute ID can be obtained too.
There is no need to go through datasets.

Also, the key issue with this implementation is that tasks of roles (e.g., aggregator) that `isDataConsumer` flag is unset won't be notified at all.

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -52,6 +55,37 @@ func NewDatasetsApiService(dbService database.DBService) openapi.DatasetsApiServ
 // CreateDataset - Create meta info for a new dataset.
 func (s *DatasetsApiService) CreateDataset(ctx context.Context, user string,
 	datasetInfo openapi.DatasetInfo) (openapi.ImplResponse, error) {
+	// check dataset realm and identify if it is computeId or region
+	datasetRealm := datasetInfo.Realm
+	isRegion := strings.HasPrefix(datasetRealm, "default/")
```
myungjin, line 60: what if datasetRealm is just "default"?

Also, define a constant, say `RootRealm` or `DefaultRealm` instead of hardcoded value.

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -52,6 +55,37 @@ func NewDatasetsApiService(dbService database.DBService) openapi.DatasetsApiServ
...
+	// If realm refers to region, check for matching computes in the region
+	// If realm refers to computeId, check if computeId exists and get its region
+	if isRegion {
+		eligibleComputes, err := s.dbService.CheckComputeByRegion(datasetRealm)
+		if err != nil {
+			return openapi.Response(http.StatusInternalServerError, nil), fmt.Errorf("failed to create new dataset: %v", err)
+		}
+		zap.S().Infof("found compute: %s for dataset in region %v", eligibleComputes[0], datasetRealm)
+		// perform realm expansion by appending with computeId
+		datasetInfo.Realm = datasetRealm + "/" + eligibleComputes[0]
```
myungjin, line 71: use path.join instead of string concatenation.

See example: https://pkg.go.dev/path#example-Join

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -52,6 +55,37 @@ func NewDatasetsApiService(dbService database.DBService) openapi.DatasetsApiServ
...
+			return openapi.Response(http.StatusInternalServerError, nil),
+				fmt.Errorf("failed to create new dataset: %v, checkResult: %v", err, checkResult)
+		}
+
+		// if computeId is found, get its region for realm expansion
+		computeRegion, err := s.dbService.GetComputeRegion(datasetRealm)
+		if err != nil {
+			return openapi.Response(http.StatusInternalServerError, nil), fmt.Errorf("failed to create new dataset: %v", err)
+		}
+		datasetInfo.Realm = computeRegion + "/" + datasetRealm
```
myungjin, line 84: same here. use path.join

myungjin (changes requested):
Left several comments.Two independent things handled in one PR. Since compute notify hardcoding fix is incorrect, it is better to focus on dataset creation check and to remove compute notify hardcoding fix.

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -52,6 +57,47 @@ func NewDatasetsApiService(dbService database.DBService) openapi.DatasetsApiServ
...
+	// Public dataset: 	The realm may not map correctly to a region or compute. For a valid datasetInfo.Realm, expansion will take place.
+	// 					For invalid datasetInfo.Realm, the datasetInfo.Realm is retained (for groupBy) but computeId will be defaultRealm
+	// Private dataset: The realm must map to a correct region or registered compute. For a valid datasetInfo.Realm, expansion will take place.
+
+	if isRegion {
+		eligibleComputes, err := s.dbService.GetComputesByRegion(datasetRealm)
+		if err != nil {
+			return openapi.Response(http.StatusInternalServerError, nil), fmt.Errorf("failed to create new dataset: %v", err)
+		}
+		if len(eligibleComputes) == 0 {
```
myungjin, line 75: If no document is found in mongodb (from the call of `GetComputesByRegion`), I thought mongodb raises an error of mongo.ErrNoDocuments. Do you assume that there can be a case where going through a cursor ended up with empty list?

Also, if mongo.ErrNoDocuments is raised, this line (line 75) will never be executed. Is this what you expect?

If you think empty list is an error case, perhaps it's better to handle it within GetComputesByRegion and here you should skip the checking the size of the list.

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -52,6 +57,47 @@ func NewDatasetsApiService(dbService database.DBService) openapi.DatasetsApiServ
...
+	datasetRealm := datasetInfo.Realm
+	isRegion := strings.HasPrefix(datasetRealm, util.DefaultRealm+"/")
+	// TODO enforce compute registration starting with defaultRealm by pre-pending region.
+	// Might move defaultRealm const to another file or utils later since it will also be used for compute registration.
+
+	// Public dataset: 	The realm may not map correctly to a region or compute. For a valid datasetInfo.Realm, expansion will take place.
+	// 					For invalid datasetInfo.Realm, the datasetInfo.Realm is retained (for groupBy) but computeId will be defaultRealm
+	// Private dataset: The realm must map to a correct region or registered compute. For a valid datasetInfo.Realm, expansion will take place.
+
+	if isRegion {
```
myungjin, line 70: Overall, the section of `if isRegion else` is too complex since it has three nested `if` statements. One way to reduce the complexity is to further introduce sub-function(s).

If you introduce sub-functions, please also consider to write some unit test cases..

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -52,6 +57,47 @@ func NewDatasetsApiService(dbService database.DBService) openapi.DatasetsApiServ
 // CreateDataset - Create meta info for a new dataset.
 func (s *DatasetsApiService) CreateDataset(ctx context.Context, user string,
 	datasetInfo openapi.DatasetInfo) (openapi.ImplResponse, error) {
+	// get datasetRealm field and identify if it is computeId or region
+	datasetRealm := datasetInfo.Realm
+	isRegion := strings.HasPrefix(datasetRealm, util.DefaultRealm+"/")
+	// TODO enforce compute registration starting with defaultRealm by pre-pending region.
+	// Might move defaultRealm const to another file or utils later since it will also be used for compute registration.
```
myungjin, line 64: Looks like you already did so with util.DefaultRealm. Please remove this comment.

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -52,6 +57,47 @@ func NewDatasetsApiService(dbService database.DBService) openapi.DatasetsApiServ
...
+	// 					For invalid datasetInfo.Realm, the datasetInfo.Realm is retained (for groupBy) but computeId will be defaultRealm
+	// Private dataset: The realm must map to a correct region or registered compute. For a valid datasetInfo.Realm, expansion will take place.
+
+	if isRegion {
+		eligibleComputes, err := s.dbService.GetComputesByRegion(datasetRealm)
+		if err != nil {
+			return openapi.Response(http.StatusInternalServerError, nil), fmt.Errorf("failed to create new dataset: %v", err)
+		}
+		if len(eligibleComputes) == 0 {
+			// No compute found for the given region. If dataset is public, set compute to DefaultRealm
```
myungjin, line 76: Where do you set DefaultRealm if dataset is public? I couldn't find any line to do so. Am I missing something here?

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -52,6 +57,47 @@ func NewDatasetsApiService(dbService database.DBService) openapi.DatasetsApiServ
...
+			}
+		} else {
+			zap.S().Infof("found compute: %s for dataset in region %v", eligibleComputes[0], datasetRealm)
+			// if region is found for public or private dataset, perform realm expansion
+			datasetInfo.Realm = path.Join(datasetRealm, eligibleComputes[0])
+		}
+	} else {
+		computeSpec, err := s.dbService.GetComputeById(datasetRealm)
+		if err != nil {
+			// No compute found for given computeId. If dataset is public, set compute to DefaultRealm
```
myungjin, line 88: Same here. Where do you set DefaultRealm in case dataset is public?

myungjin (commented):
comments are left. Please consider to address them.

pkg/openapi/controller/api_datasets_service.go
dhruvsgarg: Okay right, there was a gap in the previous implementation. For public datasets, if the realm doesnt match the compute infra, we want to retain the dataset's realm (for our TAG) but assign it to a default compute. I have updated the code to use a `computeFound` bool variable in the `DatasetInfo` component. This variable will be true if a matching computeId OR a computeId in the given region is found. Thus, when creating a task, we check if computeFound is true. If it is not true, we assign it the defaultRealm's compute.

pkg/openapi/controller/api_datasets_service.go
dhruvsgarg: I have updated the implementation to return an empty string list from `GetComputeIdsByRegion` if there is any mongodb error in fetching documents. Hence, if ErrNoDocuments is encountered, an empty string list will be returned. along with the error message. This way we can just check for computes by region once, and assign compute accordingly.

pkg/openapi/controller/api_datasets_service.go
dhruvsgarg: I created the functions but couldn't write test cases since these functions also use the DBService. If you suggest a way, I can include these test cases in the next PR

myungjin: One way is to create a mock interface of DBService; let's not worry about it here as the sub-functions look much clearer.

myungjin (commented):
the revision is almost there. please consider to address a few comments.

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -110,3 +141,41 @@ func (s *DatasetsApiService) UpdateDataset(ctx context.Context, user string, dat
 
 	return openapi.Response(http.StatusNotImplemented, nil), errors.New("UpdateDataset method not implemented")
 }
+
+func (s *DatasetsApiService) checkComputeForDatasetByRegion(datasetRealm string, datasetInfo *openapi.DatasetInfo) error {
+	eligibleComputes, err := s.dbService.GetComputeIdsByRegion(datasetRealm)
```
myungjin, line 146: There is no check on `err`. Is this safe or can the error be ignored? If so, you may need to change it to something like the following:

```
eligibleComputes, _ := s.dbService.GetComputeIdsByRegion(datasetRealm)
```

myungjin (approved):
one minor issue to think about. But overall lgtm.

PR #202, opened 2022-08-04
Automated deployment notification to computes

dhruvsgarg:
A job consists of several tasks that might be deployed across different compute clusters. This pr allows the controller to automatically identify and notify a list of computes for deploying resources for tasks.

### DIFF
--- a/cmd/controller/app/job/default_handler.go
+++ b/cmd/controller/app/job/default_handler.go
@@ -323,8 +323,22 @@ func (h *DefaultHandler) notifyJob(evtType pbNotify.JobEventType) error {
 }
 
 func (h *DefaultHandler) notifyDeploy(evtType pbNotify.DeployEventType) error {
-	// TODO: add computeId field to tasks and remove hard-coding
-	computeIds := []string{"compute-1"}
+	// Get a list of computeIds to notify for deployment
+	// To do this, get all tasks for the jobId, iterate and get computeId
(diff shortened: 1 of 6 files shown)


cmd/controller/app/job/default_handler.go
```diff
@@ -323,8 +323,28 @@ func (h *DefaultHandler) notifyJob(evtType pbNotify.JobEventType) error {
 }
 
 func (h *DefaultHandler) notifyDeploy(evtType pbNotify.DeployEventType) error {
-	// TODO: add computeId field to tasks and remove hard-coding
-	computeIds := []string{"compute-1"}
+	// Get a list of computeIds to notify for deployment
+	// To do this, get all tasks for the jobId, iterate and get computeId
+	computeIdMap := make(map[string]bool)
+
+	for _, taskInfo := range h.tasksInfo {
```
myungjin, line 330: What if we define an attribute of ComputeId in TaskInfo and GetTasksInfo() and GetTaskInfo() return TaskInfo struct or array of the struct?

Then, I don't think that GetSelectTaskProperties() are not needed. I feel that introducing GetSelectTaskProperties() incurs too many changes.

myungjin (commented):
left a comment. Please consider if augmenting TaskInfo helps to keep code simple.

myungjin (approved):
lgtm

PR #205, opened 2022-08-08
Duplicating controllers deployment code into deployer

dhruvsgarg:
This pr is a step towards implementing the core deployment logic into the deployer. We duplicate the relevant code and logic from the controller's deployer and put it as placeholder in the standalone deployer's code. At a later stage, only logic within the deployer will remain and all deployment related code from the controller will be removed.

### DIFF
--- a/cmd/deployer/app/resource_handler.go
+++ b/cmd/deployer/app/resource_handler.go
@@ -172,9 +172,27 @@ func (r *resourceHandler) addResource(jobId string) {
 	zap.S().Infof("Successfully added resources for compute %s and jobId %s", r.spec.ComputeId, jobId)
 }
 
-func (r *resourceHandler) revokeResource(jobId string) {
+func (r *resourceHandler) revokeResource(jobId string) error {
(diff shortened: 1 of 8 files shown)


cmd/deployer/app/resource_handler.go
```diff
@@ -172,9 +172,27 @@ func (r *resourceHandler) addResource(jobId string) {
 	zap.S().Infof("Successfully added resources for compute %s and jobId %s", r.spec.ComputeId, jobId)
 }
 
-func (r *resourceHandler) revokeResource(jobId string) {
+func (r *resourceHandler) revokeResource(jobId string) error {
 	zap.S().Infof("Received revoke resource request for job %s", jobId)
-	// TODO: implement revokeResource method
+
+	// WORK IN PROGRESS: Moved code directly from controller. Contains only core logic which could be used.
```
myungjin, line 178: One minor point to think about: Commenting these lines won't keep a next PR small if you have to uncomment and/or rewrite them.

myungjin (approved):
overall looks good. But left one comment for consideration.

PR #208, opened 2022-08-09
Enabled deployment of resources by computes

dhruvsgarg:
This pr removes the deployer package within the controller. The deployment of resources for tasks is now done by deployers of various compute clusters.


cmd/controller/app/job/default_handler.go
```diff
@@ -198,17 +187,6 @@ func (h *DefaultHandler) cleanup() {
...
-		if err := h.dplyr.Uninstall("job-" + h.jobId); err != nil {
-			zap.S().Warnf("failed to release resources for job %s: %v", h.jobId, err)
-		}
-	}
-
-	// 2.delete all the job resource specification files
-	deploymentChartPath := filepath.Join(deploymentDirPath, h.jobId)
-	_ = os.RemoveAll(deploymentChartPath)
-
 	h.tskWatchCancelFn()
```
dhruvsgarg, line 190: Before invoking `taskWatchCancelFn()` do we check here if the deployers have actually uninstalled the resources? If so, it might require a separate API call right?

cmd/deployer/app/resource_handler.go
```diff
@@ -175,19 +202,16 @@ func (r *resourceHandler) addResource(jobId string) {
...
+	}
 
-	// // 2.delete all the job resource specification files
-	// deploymentChartPath := filepath.Join(deploymentDirPath, h.jobId)
-	// _ = os.RemoveAll(deploymentChartPath)
+	// 2.delete all the job resource specification files
+	deploymentChartPath := filepath.Join(deploymentDirPath, jobId)
+	_ = os.RemoveAll(deploymentChartPath)
 
 	// h.tskWatchCancelFn()
```
dhruvsgarg, line 216: I have left this here to confirm with you that the task watch cancel function is for the controller itself and not the deployer. Will remove it once you confirm

cmd/deployer/app/resource_handler.go
```diff
@@ -219,91 +243,80 @@ func (r *resourceHandler) getDeploymentConfig(jobId string) (openapi.DeploymentC
...
+		dstFilePath := filepath.Join(deploymentChartPath, chartFile)
+		if err := util.CopyFile(srcFilePath, dstFilePath); err != nil {
+			errMsg := fmt.Sprintf("failed to copy a deployment chart file %s: %v", chartFile, err)
+			zap.S().Debugf(errMsg)
+			return fmt.Errorf(errMsg)
+		}
+	}
+
+	for _, agentKV := range deploymentConfig.AgentKVs {
+		// the agentKV is a map but with just one key (taskId) and one value (taskKey)
```
dhruvsgarg, line 266: The agentKV is a list of maps with [taskID:taskKey]. However each map item has only one key-value pair. This can be updated in a separate pr later if need be

cmd/deployer/app/resource_handler.go
```diff
@@ -219,91 +243,80 @@ func (r *resourceHandler) getDeploymentConfig(jobId string) (openapi.DeploymentC
...
+		deploymentFilePath := filepath.Join(targetTemplateDirPath, deploymentFileName)
+		err = ioutil.WriteFile(deploymentFilePath, []byte(rendered), util.FilePerm0644)
+		if err != nil {
+			errMsg := fmt.Sprintf("failed to write a job rosource spec %s: %v", taskId, err)
+			zap.S().Debugf(errMsg)
+			return fmt.Errorf(errMsg)
+		}
+	}
+
+	// TODO: when multiple clusters are supported,
```
dhruvsgarg, line 298: I think this (old todo) can be removed now. We will be able to have different clusters supporting different backends right?

cmd/deployer/cmd/root.go
```diff
@@ -139,6 +151,14 @@ func init() {
 	rootCmd.Flags().StringP(argApiKey, "k", defaultApiKey, "unique apikey")
 	rootCmd.MarkFlagRequired(argApiKey)
 
+	defaultPlatform := "k8s"
+	rootCmd.Flags().StringP(argPlatform, "p", defaultPlatform, "compute platform")
+	rootCmd.MarkFlagRequired(argPlatform)
+
+	defaultNamespace := "flame"
```
dhruvsgarg, line 158: I am assuming here that it is okay for the deployer to take the namespace as an argument

cmd/deployer/cmd/root.go
```diff
@@ -139,6 +151,14 @@ func init() {
 	rootCmd.Flags().StringP(argApiKey, "k", defaultApiKey, "unique apikey")
 	rootCmd.MarkFlagRequired(argApiKey)
 
+	defaultPlatform := "k8s"
```
dhruvsgarg, line 154: Does a platform field need to be added in the computeCollection in the mongodb?

cmd/controller/app/job/default_handler.go
```diff
@@ -221,86 +199,6 @@ func (h *DefaultHandler) allocateComputes() error {
 	h.notifyDeploy(pbNotify.DeployEventType_ADD_RESOURCE)
```
myungjin, line 199: Similar to the deallocation situation, we need to check if resources are allocated or not. If not, we can't make a transition to a different state. Add TODO here and handle this issue in a next PR.

cmd/deployer/app/resource_handler.go
myungjin: You're right. That function can't be called here. So, please remove it.

cmd/deployer/app/resource_handler.go
myungjin: Yes, that's an old TODO. you can remove it.

cmd/deployer/app/resource_handler.go
myungjin: Actually, the definition of api specification is wrong. We don't need a list of maps; what is needed is a simple map. Please address it in a next PR.

cmd/deployer/app/resource_handler.go
```diff
@@ -219,91 +243,80 @@ func (r *resourceHandler) getDeploymentConfig(jobId string) (openapi.DeploymentC
 }
 
 func (r *resourceHandler) deployResources(deploymentConfig openapi.DeploymentConfig) error {
```
myungjin, line 245: The caller of this function needs to update deployment status to the DB via restapi. Please add TODO in an appropriate location and work on it in a next PR.

cmd/deployer/app/resource_handler.go
```diff
@@ -219,91 +243,80 @@ func (r *resourceHandler) getDeploymentConfig(jobId string) (openapi.DeploymentC
...
+		return fmt.Errorf(errMsg)
+	}
+
+	if err = dplyr.Initialize("", r.namespace); err != nil {
+		errMsg := fmt.Sprintf("failed to initialize a job deployer: %v", err)
+		zap.S().Debugf(errMsg)
+		return fmt.Errorf(errMsg)
+	}
+
+	err = dplyr.Install("job-"+deploymentConfig.JobId+"-"+r.spec.ComputeId, deploymentChartPath)
```
myungjin, line 313: I am not certain why you need ComputeId as helm chart installation name. Every helm chart installed in one deployer has the same compute id. So, as long as job id is unique, compute id doesn't contribute to creating a unique name.

Unless there is a strong reason, please remove ComputeId.

cmd/deployer/cmd/root.go
myungjin: not necessarily because we only support k8s at the moment.
When we actually support another platform (docker is a placeholder), we may need to create platform field at that point. But not right now.

cmd/deployer/cmd/root.go
myungjin: that's fine.

fiab/helm-chart/templates/controller-configmap.yaml
```diff
@@ -25,7 +25,6 @@ data:
     ---
     db: mongodb://flame-mongodb-headless:27017/?replicaSet=rs0
     notifier: {{ .Values.frontDoorUrl.notifier }}:443
-    platform: k8s
     namespace: {{ .Release.Namespace }}
```
myungjin, line 28: Why do you keep `namespace` variable as controller no longer uses namespace variable (see https://github.com/cisco-open/flame/pull/208/files#diff-fead7ba084b7aea2bfd0b0655b451b196d391d6c4afb2cc6fbebaa2ea08c8816L32-L33)?

Please remove this line if it's no longer needed.

myungjin (commented):
left several comments. Mostly they are minor ones. Please consider to address them.

cmd/deployer/app/resource_handler.go
dhruvsgarg: I have kept compute-id in the helm chart installation to support deployment of multiple clusters on a local system. This is because if deployer-default uses job-<job_id> as a release name to create some pods, then this same release name cannot be used to create pods for tasks in the same job_id. Thus, I changed it to job-<job_id>-<computeId> so that it works on both local and actual multi-cluster settings

myungjin (approved):
lgtm

PR #211, opened 2022-08-11
Removed deprecated io-ioutil API calls

dhruvsgarg:
The golang linter complained that io-ioutil is deprecated. Thus the function calls were replaced with relevant io and os package APIs.


myungjin (approved):
lgtm

PR #213, opened 2022-08-12
Initial error handling for deployer

dhruvsgarg:
This pr adds initial error handling for the deployer. Failures at the controller while sending events to the deployer and at the deployer while adding or revoking resources can be handled. Propagating the error back from deployer to controller will be a part of another pr.


myungjin (approved):
lgtm

PR #237, opened 2022-09-28
event driven approach

myungjin:
The flame library presents some inefficiency.

1) In order to check if there is at least one peer in a channel, it busy-waits by calling empty() function and sleep one second.

2) head-of-line blocking problem: to recv data from peers of a channel, the implementation gets the list of peers, loops through each of peers and calls recv(). If the first peer is a straggler, it causes a head-of-line blocking problem.

3) When model size increases, it takes a while to merge transmitted messages (it is byte array concatenation operation). This process takes place in a coroutine; and it isn't suspended until it is finished. This delays a heart beat transmission as the heart beat function is also a coroutine. The delayed heart beat triggers a timeout at the other side of grpc channel, which breaks a training job.

The first two issues are addressed in an event-driven manner. The third problem is addressed by using threading. The byte array concatenation takes place in a separate thread. So, the asyncio coroutine of handling message will release cpu resource quickly. Therefore, the heart beat transmission function works in a timely manner.

Also, graceful termination is enforced by ensuring data transfer is complete.

### DIFF
--- a/lib/python/flame/backend/p2p.py
+++ b/lib/python/flame/backend/p2p.py
@@ -19,12 +19,12 @@ import asyncio
 import logging
 import socket
 import time
-from typing import AsyncIterable, Iterable, Tuple
+from typing import AsyncIterable, Iterable
 
 import grpc
 
-from ..common.constants import (DEFAULT_RUN_ASYNC_WAIT_TIME, BackendEvent,
-                                CommType)
+from ..common.constants import (DEFAULT_RUN_ASYNC_WAIT_TIME, EMPTY_PAYLOAD,
+                                BackendEvent, CommType)
 from ..common.util import background_thread_loop, run_async
 from ..proto import backend_msg_pb2 as msg_pb2
 from ..proto import backend_msg_pb2_grpc as msg_pb2_grpc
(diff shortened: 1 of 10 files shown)


lib/python/flame/backend/p2p.py
```diff
@@ -72,15 +74,15 @@ async def recv_data(self, req: msg_pb2.BackendID,
...
         dck_task = asyncio.create_task(self._dummy_context_keeper(context))
-        self.p2pbe._set_writer(req.end_id, context)
+        await self.p2pbe._set_writer(req.end_id, context)
         self.p2pbe._set_heart_beat(req.end_id)
 
         await dck_task
 
     async def _dummy_context_keeper(self, context):
-        """Sleep 1000 sec (an abitrary big value) to keep context alive."""
+        """Sleep 100000 sec (an abitrary big value) to keep context alive."""
```
RaviKhandavilli, line 83: check if there is any method , as long sleeps does not look good

lib/python/flame/common/constants.py
```diff
@@ -27,6 +27,8 @@
 # default data folder
 DATA_FOLDER_PATH = '/flame/data'
 
+EMPTY_PAYLOAD = b''
```
RaviKhandavilli, line 30: check syntax

RaviKhandavilli (approved):
LGTM

lib/python/flame/backend/p2p.py
myungjin: That is a good suggestion. I will address your comment.

lib/python/flame/common/constants.py
myungjin: I think this syntax is correct. can you elaborate it if I miss something?

RaviKhandavilli: I feel like b should be inside quotes

RaviKhandavilli (approved):
LGTM

PR #245, opened 2022-10-09
delete design code

RaviKhandavilli:
Added code to flamectl cli to Support remove command.
Added code to process remove command Http request on Api server and controller.
Added code to update DB , when a code is removed.

### DIFF
--- a/cmd/controller/app/database/mongodb/code.go
+++ b/cmd/controller/app/database/mongodb/code.go
@@ -202,3 +202,22 @@ func (db *MongoService) GetDesignCode(userId string, designId string, fileVer st
 
 	return util.ZipFile(updatedDoc.DataSet)
 }
+func (db *MongoService) DeleteDesignCode(userId string, designId string, version string) error {
+	zap.S().Debugf("delete design code 1: %v, %v,%v", userId, designId, version)
+
+	if version == latestVersion {
+		version = db.GetLatestDesignCodeVersion(userId, designId)
+	}
+	updateRes, err := db.designCollection.UpdateOne(context.TODO(),
+		bson.M{util.DBFieldUserId: userId, util.DBFieldId: designId},
+		bson.M{"$pull": bson.M{util.DBFieldCodes: bson.M{"version": version}}})
+	if err != nil {
+		return fmt.Errorf("failed to delete design code deleted code error: %v", err)
+	}
+	if updateRes.ModifiedCount == 0 {
+		return fmt.Errorf("failed to delete design code, code version %s not found. deleted code count: %#v",
+			version, updateRes)
+	}
+	zap.S().Debugf("successfully deleted design code: %#v", updateRes)
+	return nil
+}
--- a/cmd/controller/app/database/mongodb/code_test.go
+++ b/cmd/controller/app/database/mongodb/code_test.go
@@ -1,0 +1,33 @@
+package mongodb
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo/integration/mtest"
+)
+
+func TestMongoService_DeleteDesignCode(t *testing.T) {
+	mt := mtest.New(t, mtest.NewOptions().ClientType(mtest.Mock))
+	defer mt.Close()
+	mt.Run("success", func(mt *mtest.T) {
+		db := &MongoService{
+			designCollection: mt.Coll,
+		}
+		mt.AddMockResponses(bson.D{{"ok", 1}, {"acknowledged", true},
+			{"n", 1}, {"nModified", 1}})
+		err := db.DeleteDesignCode("userid", "designid", "version")
+		assert.Nil(t, err)
+	})
+
+	mt.Run("no document deleted", func(mt *mtest.T) {
+		db := &MongoService{
+			designCollection: mt.Coll,
+		}
+		mt.AddMockResponses(bson.D{{"ok", 1}, {"acknowledged", true},
+			{"n", 0}, {"nModified", 0}})
+		err := db.DeleteDesignCode("userid", "designid", "version")
+		assert.NotNil(t, err)
+	})
+}
--- a/cmd/controller/cmd/root.go
+++ b/cmd/controller/cmd/root.go
@@ -53,7 +53,7 @@ var (
 
 			instance, err := app.NewController(cfg, bInsecure, bPlain)
 			if err != nil {
-				fmt.Printf("Failed to create a new controller instance\n")
+				fmt.Printf("Failed to create a new controller instance: %v\n", err)
 				return nil
 			}
 			instance.Start()
--- a/cmd/flamectl/cmd/delete_code.go
+++ b/cmd/flamectl/cmd/delete_code.go
@@ -1,0 +1,55 @@
+// Copyright 2022 Cisco Systems, Inc. and its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/cisco-open/flame/cmd/flamectl/resources/code"
+)
+
+var removeCodeCmd = &cobra.Command{
+	Use:   "code <version> --design <design id>",
+	Short: "Remove a code with specific version from a design",
+	Long:  "This command removes a code with specific version from a design",
+	Args:  cobra.RangeArgs(1, 1),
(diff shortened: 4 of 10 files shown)


cmd/controller/app/database/mongodb/code.go
```diff
@@ -202,3 +202,22 @@ func (db *MongoService) GetDesignCode(userId string, designId string, fileVer st
...
 }
+func (db *MongoService) DeleteDesignCode(userId string, designId string, version string) error {
+	zap.S().Debugf("delete design code 1: %v, %v,%v", userId, designId, version)
+
+	if version == "" || version == latestVersion {
+		version = db.GetLatestDesignCodeVersion(userId, designId)
+	}
+	updateRes, err := db.designCollection.UpdateOne(context.TODO(),
+		bson.M{util.DBFieldUserId: userId, util.DBFieldId: designId},
+		bson.M{"$pull": bson.M{"codes": bson.M{"version": version}}})
```
myungjin, line 213: Can we try to make the key word, "codes", as a constant? In this file, there are a lot of places that use that key word. It would be nice to make it a constant so that we can prevent future tech debts?

We can handle this in a separate PR.

cmd/controller/app/database/mongodb/code_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo/integration/mtest"
+)
+
+func TestMongoService_DeleteDesignCode(t *testing.T) {
+	mt := mtest.New(t, mtest.NewOptions().ClientType(mtest.Mock))
+	defer mt.Close()
+	mt.Run("success", func(mt *mtest.T) {
```
myungjin, line 14: Is this possible to add more than one code versions and delete only one of them and check if the specified version is only removed?

cmd/controller/cmd/root.go
```diff
@@ -53,7 +53,7 @@ var (
 
 			instance, err := app.NewController(cfg, bInsecure, bPlain)
 			if err != nil {
-				fmt.Printf("Failed to create a new controller instance\n")
+				fmt.Printf("Failed to create a new controller instance:%v\n", err)
```
myungjin, line 56: nit
```suggestion
				fmt.Printf("Failed to create a new controller instance: %v\n", err)
```

cmd/flamectl/cmd/delete_code.go
```diff
@@ -0,0 +1,54 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"github.com/cisco-open/flame/cmd/flamectl/resources/code"
+	"github.com/spf13/cobra"
```
myungjin, line 20: Please look at the 2nd bullet point of https://github.com/cisco-open/flame/blob/main/CONTRIBUTING.md#use-a-consistent-coding-style and group packages accordingly.

cmd/flamectl/cmd/delete_code.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+var removeCodeCmd = &cobra.Command{
+	Use:   "code <version>",
```
myungjin, line 24: Is this usage format correct? What about `-d` argument? Can you double-check if this format is correct?

cmd/flamectl/cmd/delete_code.go
```diff
@@ -0,0 +1,54 @@
...
+	Short: "Remove a code",
+	Long:  "This command removes a code",
```
myungjin, line 26: Can you add more description? A code should be associated with a design. So, a code should be removed from a design. Can you make the description a bit more informative?

cmd/flamectl/resources/code/code.go
```diff
@@ -72,7 +72,7 @@ func Create(params Params) error {
 		return nil
 	}
 
-	fmt.Printf("Code created successfully for design %s\n", params.DesignId)
+	fmt.Printf("Code created successfully for design %s, version %v\n", params.DesignId, params.CodeVer)
```
myungjin, line 75: When a code is created, user doesn't need to provide CodeVer. The DB code for creating a code will determine a version (https://github.com/cisco-open/flame/pull/245/files#diff-2164dd2c9fd70fa0f3feb28786e75a81fbe9feed581c045325da3c98c5a331e0R46-R47). So, here params.CodeVer will always print 0 or some arbitrary value.

We first need to check what response body returns and make sure that a code version is in the response body.

We may have to enhance this with a separate PR.

cmd/controller/app/database/mongodb/code.go
```diff
@@ -202,3 +202,22 @@ func (db *MongoService) GetDesignCode(userId string, designId string, fileVer st
 
 	return util.ZipFile(updatedDoc.DataSet)
 }
+func (db *MongoService) DeleteDesignCode(userId string, designId string, version string) error {
+	zap.S().Debugf("delete design code 1: %v, %v,%v", userId, designId, version)
+
+	if version == "" || version == latestVersion {
```
myungjin, line 208: Since it involves deletion of resource, we may want to be conservative. Perhaps would it be better to return an error if a version string is empty?

myungjin (commented):
Left several comments. Please consider to revise them.

cmd/flamectl/resources/code/code.go
```diff
@@ -112,3 +112,23 @@ func mustOpen(f string) *os.File {
 
 	return r
 }
+func Remove(params Params) error {
```
myungjin, line 115: Can you also unit-test this code as well?

cmd/controller/app/database/mongodb/code.go
```diff
@@ -202,3 +202,22 @@ func (db *MongoService) GetDesignCode(userId string, designId string, fileVer st
...
+		bson.M{util.DBFieldUserId: userId, util.DBFieldId: designId},
+		bson.M{"$pull": bson.M{"codes": bson.M{"version": version}}})
+	if err != nil {
+		return fmt.Errorf("failed to delete design code deleted code error: %v", err)
+	}
+	if updateRes.ModifiedCount == 0 {
+		return fmt.Errorf("failed to delete design code, code version %s not found. deleted code count: %#v",
+			version, updateRes)
+	}
+	zap.S().Debugf("successfully deleted design code:%#v", updateRes)
```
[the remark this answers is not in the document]
RaviKhandavilli: done

cmd/controller/app/database/mongodb/code.go
RaviKhandavilli: For now I have created a constant and used it in this code . we will plan and replace similar strings in different PR

cmd/controller/app/database/mongodb/code_test.go
RaviKhandavilli: we have just mocked mongo db responses here, I suggest we should do this as part of integration test.

cmd/controller/cmd/root.go
RaviKhandavilli: done

cmd/flamectl/cmd/delete_code.go
RaviKhandavilli: here both are third party packages and grouped together , I don't think new line is necessary.

cmd/flamectl/cmd/delete_code.go
RaviKhandavilli: -d argument was made as mandate MarkPersistentFlagRequired, so command will fail without this argument. Anyways added it in description.

cmd/flamectl/cmd/delete_code.go
RaviKhandavilli: done

cmd/flamectl/resources/code/code.go
RaviKhandavilli: we need to correct logs of flamectl in few other locations as well, we shall handle it is a part of different PR, for now removing this change.

cmd/controller/app/database/mongodb/code.go
RaviKhandavilli: done.

cmd/flamectl/resources/code/code.go
RaviKhandavilli: To write unit cases for this function we need to refactor restapi package.as this function has dependencies with functions in restapi Package.

cmd/flamectl/cmd/delete_code.go
myungjin: The first one is ours, not a 3rd party one. My recommendation is:
```
	"github.com/spf13/cobra"

	"github.com/cisco-open/flame/cmd/flamectl/resources/code"
```

cmd/flamectl/cmd/delete_code.go
myungjin: Thanks for this. I was asking if someone runs `flamectl remove code -h`, what would be a usage output?
I was asking if the usage is correctly printed out. Should `-d` be part of a usage description as it is mandatory?

cmd/flamectl/resources/code/code.go
myungjin: Sounds good.

cmd/flamectl/cmd/delete_code.go
RaviKhandavilli: Done

myungjin (approved):
lgtm

PR #261, opened 2022-11-03
feat: 'continue' primitive in tasklet

myungjin:
We add 'continue' primitive in the tasklet code of the flame library. The 'continue' primitive is used to move back to the start of loop when a certain condition is met. A condition is defined as a function, which will be set via set_continue_fn member function.

As an example use case of 'continue' primitive, a middle aggregator code is extended in the following way: If no trainer join a channel that the middle aggregator is waiting on for a minute, the middle aggregator gives up waiting and updates the top aggregator with dummy weights so that the top aggregator can be unblocked. Then, the middle aggregator goes back to the start of the loop so that it can retrieve model weights of a new global model from the top aggregator.

### DIFF
--- a/lib/python/flame/channel.py
+++ b/lib/python/flame/channel.py
@@ -275,15 +275,31 @@ class Channel(object):
         """Join the channel."""
         self._backend.join(self)
 
-    def await_join(self):
-        """Wait for at least one peer joins a channel."""
+    def await_join(self, timeout=None) -> bool:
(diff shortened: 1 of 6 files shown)


harshitdaga (approved):
lgtm!

RaviKhandavilli (approved):
LGTM

Divlik (approved):

PR #262, opened 2022-11-07
deployment status

RaviKhandavilli:
1.Code added to forward deployment status to controller .
2.Changed deployer logic to deploy each task separately so as to update task status of individual task.
3.Changed release name of the agents to contain job id and task id (only first 12 characters of task id ) in the place of job id and compute id.
4.Made some refactoring changes to use global constants instead of local strings.
Note :- Deployment status only represents whether the job added is deployed/failed/revoked. it does not contain task status.task status is directly communicated back to controller by flamelet agents.
DB changes :- 
1.created a separate collection deployment to mongodb to specifically maintain deployment status of computes.


cmd/controller/app/database/mongodb/deployment.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func (db *MongoService) UpdateDeploymentStatus(deploymentStatus openapi.DeploymentStatus) error {
+	filter := bson.M{util.DBFieldJobId: deploymentStatus.JobId, util.DBFieldComputeId: deploymentStatus.ComputeId}
+	result := db.deploymentCollection.FindOne(context.TODO(), filter)
+	// update fields if document is already present
+	if result.Err() == nil {
+		taskStatus := map[string]openapi.AgentState{}
+		for key, val := range deploymentStatus.AgentStatuses {
+			taskStatus["agentstatuses."+key] = val
```
myungjin, line 20: can you make `agentstatuses` as a constant?

cmd/controller/app/database/mongodb/deployment.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.uber.org/zap"
+)
+
+func (db *MongoService) UpdateDeploymentStatus(deploymentStatus openapi.DeploymentStatus) error {
+	filter := bson.M{util.DBFieldJobId: deploymentStatus.JobId, util.DBFieldComputeId: deploymentStatus.ComputeId}
+	result := db.deploymentCollection.FindOne(context.TODO(), filter)
+	// update fields if document is already present
+	if result.Err() == nil {
```
myungjin, line 17: There is a chance that a set of agents in a deploymentStatus change over time.
Say, in one execution: the set has {a, b, c} and in a future execution, it may be {b, c, d}.

In that case, we need to remove {a} from the db and add {d} into the db while updating the status of {b, c}.

This code can't handle removal. Can you please handle that?

cmd/controller/app/database/mongodb/service.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/cisco-open/flame/pkg/util"
```
myungjin, line 23: place it after line 28 with a new line before it.

cmd/deployer/app/resource_handler.go
```diff
@@ -221,23 +228,57 @@ func (r *resourceHandler) addResource(jobId string) error {
...
-			return err
+	cfg, err := r.getDeploymentConfig(jobId)
+	if err != nil {
+		zap.S().Errorf("Failed to get %s job configs", jobId)
+		return fmt.Errorf("failed to revoke %s job: %w", jobId, err)
+	}
+	for taskId := range cfg.AgentKVs {
+		// 1. decommission compute resources one by one if they are in use
+		if r.dplyr != nil {
+			if revokeErr := r.dplyr.Uninstall("job-" + jobId + "-" + taskId[:12]); err != nil {
```
myungjin, line 239: this numerical value, `12` should be defined as a constant because it is used more than one place.

myungjin, line 239: what you get is `revokeErr`, but the if-statement checks `err`. This looks incorrect.

cmd/deployer/app/resource_handler.go
```diff
@@ -221,23 +228,57 @@ func (r *resourceHandler) addResource(jobId string) error {
...
+				zap.S().Warnf("failed to release resources for job %s: %v", jobId, err)
+				err = fmt.Errorf("%v: %v", err, revokeErr)
+				continue
+			}
+			r.postDeploymentStatus(jobId, taskId, openapi.AGENT_REVOKED)
```
myungjin, line 244: I think this can create too many requests and is inefficient.

Instead, you can keep the successful revoke taskIds in a map (map[string]openapi.AgentState) and pass the the map to postDeploymentStatus function.

cmd/deployer/app/resource_handler.go
```diff
@@ -221,23 +228,57 @@ func (r *resourceHandler) addResource(jobId string) error {
...
 		}
 	}
-
 	// 2.delete all the job resource specification files
 	deploymentChartPath := filepath.Join(deploymentDirPath, jobId)
 	_ = os.RemoveAll(deploymentChartPath)
+	if err != nil {
```
myungjin, line 250: `err` is a variable defined from getDeploymentConfig call. So, something doesn't seem right.

cmd/deployer/app/resource_handler.go
```diff
@@ -221,23 +228,57 @@ func (r *resourceHandler) addResource(jobId string) error {
...
+				zap.S().Warnf("failed to release resources for job %s: %v", jobId, err)
+				err = fmt.Errorf("%v: %v", err, revokeErr)
+				continue
+			}
+			r.postDeploymentStatus(jobId, taskId, openapi.AGENT_REVOKED)
 		}
 	}
-
 	// 2.delete all the job resource specification files
 	deploymentChartPath := filepath.Join(deploymentDirPath, jobId)
```
myungjin, line 248: There is no such directory that only ends with jobId because now the helm-chart directory consists of three elements: deploymentDirPath, jobId and taskId (see https://github.com/cisco-open/flame/pull/262/files#diff-9d9bde33b386c8e731e98f3cdc370505e45708cf8274a6b701ea7e58b6858353R312).

Each directory should be deleted separately only if revoke is successful. Therefore lines 248-249 should be moved into `for taskId := range cfg.AgentKVs { ... }` loop.

cmd/deployer/app/resource_handler.go
```diff
@@ -221,23 +228,57 @@ func (r *resourceHandler) addResource(jobId string) error {
...
+	// construct url
+	uriMap := map[string]string{
+		"computeId": r.spec.ComputeId,
+		"jobId":     jobId,
+	}
+
+	deploymentStatus := openapi.DeploymentStatus{
+		JobId:         jobId,
+		ComputeId:     r.spec.ComputeId,
+		AgentStatuses: map[string]openapi.AgentState{taskId: status},
```
myungjin, line 271: Please modify this function so that it can create a request for many agents instead of one agent.

cmd/deployer/app/resource_handler.go
```diff
@@ -300,25 +346,15 @@ func (r *resourceHandler) deployResources(deploymentConfig openapi.DeploymentCon
 			errMsg := fmt.Sprintf("failed to write a job rosource spec %s: %v", taskId, err)
 			return fmt.Errorf(errMsg)
 		}
+		//using short id of task as label name does not support more than 35 characters
+		err = r.dplyr.Install("job-"+deploymentConfig.JobId+"-"+taskId[:12], deploymentChartPath)
+		if err != nil {
+			errMsg := fmt.Sprintf("failed to deploy tasks: %v", err)
+			r.postDeploymentStatus(deploymentConfig.JobId, taskId, openapi.AGENT_FAILED)
```
myungjin, line 353: Can you also do this in a batch fashion so that the number requests can reduced?

cmd/deployer/app/resource_handler.go
```diff
@@ -300,25 +346,15 @@ func (r *resourceHandler) deployResources(deploymentConfig openapi.DeploymentCon
...
+			return fmt.Errorf(errMsg)
+		}
+		r.postDeploymentStatus(deploymentConfig.JobId, taskId, openapi.AGENT_DEPLOYED)
```
myungjin, line 356: Same here!

myungjin (commented):
left comments. please consider to address them.

api/computes_components.partials.yml
```diff
@@ -101,10 +101,10 @@ DeploymentStatus:
   properties:
     jobId:
       type: string
+    computeId:
```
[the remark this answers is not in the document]
RaviKhandavilli: we need this encapsulation of job id and compute id to make inserting data to MongoDB simple, also will be useful while implementing get multiple deployment status codes.
Anyways for put call I have changed message to contain only agent-statuses obj instead of deployment status obj.

cmd/controller/app/database/mongodb/deployment.go
RaviKhandavilli: In such case , we might also need to remove helm charts from deployer as well . I feel like this case might need to be dealt separately .

pkg/openapi/controller/api_computes_service.go
```diff
@@ -170,22 +170,14 @@ func (s *ComputesApiService) GetDeployments(ctx context.Context, computeId strin
...
-	//TODO: Uncomment the next line to return response Response(401, {}) or use other options such as http.Ok ...
-	//return Response(401, nil),nil
-
-	//TODO: Uncomment the next line to return response Response(0, Error{}) or use other options such as http.Ok ...
-	//return Response(0, Error{}), nil
-
-	return openapi.Response(http.StatusNotImplemented, nil), errors.New("PutDeploymentStatus method not implemented")
+func (s *ComputesApiService) PutDeploymentStatus(ctx context.Context, computeId string, jobId string, xAPIKEY string,
+	deploymentStatus openapi.DeploymentStatus) (openapi.ImplResponse, error) {
+	err := s.dbService.UpdateDeploymentStatus(deploymentStatus)
```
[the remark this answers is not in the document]
RaviKhandavilli: Changed api endpoint to send only agent statuses obj , instead of deployment statuses obj.
Constructing deployment status obj locally from both params(job id and compute id) and body.

cmd/controller/app/database/mongodb/deployment.go
RaviKhandavilli: done

cmd/controller/app/database/mongodb/service.go
RaviKhandavilli: done

cmd/deployer/app/resource_handler.go
RaviKhandavilli: Corrected

cmd/deployer/app/resource_handler.go
RaviKhandavilli: changed code to collectively post statuses.

cmd/deployer/app/resource_handler.go
RaviKhandavilli: corrected

cmd/deployer/app/resource_handler.go
RaviKhandavilli: changed it to update statuses in batch

cmd/deployer/app/resource_handler.go
```diff
@@ -219,23 +227,61 @@ func (r *resourceHandler) addResource(jobId string) error {
...
+	}
+	taskStatuses := map[string]openapi.AgentState{}
+	defer r.postDeploymentStatus(jobId, taskStatuses)
+	for taskId := range cfg.AgentKVs {
+		// 1. decommission compute resources one by one if they are in use
+		if r.dplyr != nil {
+			uninstallErr := r.dplyr.Uninstall("job-" + jobId + "-" + taskId[:k8sShortLabelLength])
+			if uninstallErr != nil {
+				zap.S().Warnf("failed to release resources for job %s: %v", jobId, uninstallErr)
+				err = fmt.Errorf("%v: %v", err, uninstallErr)
```
myungjin, line 246: This code snippet concatenates errors. The colon(:) separator makes it hard to separate out each errors when it is read by human engineers because colons can be used in an error. See an example from your PR: https://github.com/cisco-open/flame/pull/262/files#diff-9d9bde33b386c8e731e98f3cdc370505e45708cf8274a6b701ea7e58b6858353R332-R333.

So, it may be better to use some other symbol such as semi-colon.

Please take care of the similar issue in other places of this PR too.

cmd/controller/app/database/mongodb/compute.go
```diff
@@ -173,3 +173,32 @@ func (db *MongoService) GetComputeById(computeId string) (openapi.ComputeSpec, e
 	}
 	return currentDocument, nil
 }
+
+func (db *MongoService) UpdateDeploymentStatus(computeId string, jobId string, agentStatuses map[string]openapi.AgentState) error {
```
myungjin, line 177: Does this handle stale record issue correctly? If not, do you want to make it as a separate PR? If so, can you create a TODO comment in this function?

myungjin (commented):
left a few minor comments; otherwise, it looks good to me. please consider to address them. this pr will be ready in the next revision.

cmd/controller/app/database/mongodb/compute.go
RaviKhandavilli: This overwrites the existing records , we are making sure to send status of all the tasks every time we put the deployment status.
instead of deleting existing records and creating new records we are sending status of all the new records in one batch. here we are simply rewriting existing record with the new data.

cmd/deployer/app/resource_handler.go
RaviKhandavilli: done

myungjin (approved):
lgtm

myungjin (approved):
lgtm

PR #268, opened 2022-11-15
CI/CD add github publish action on release to push docker image

Divlik:
1. Adds separate workflow for docker build and publish
2. Adds job trigger only on new release tag publish
3. Adds release tag as docker image tag

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,45 @@
+name: Publish
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+
+jobs:
+
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      -
+        name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.16-alpine
```
myungjin, line 1: Do we need this? This is just an example. Now that the build/Dockerfile is working fine, I guess this is not needed. Can you share your opinion?

main.go
```diff
@@ -0,0 +1,21 @@
+package main
```
myungjin, line 1: The same comment here. Do we need this toy example? Otherwise, please remove it.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,48 @@
...
+
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
```
myungjin, line 20: Is this really needed? what is the benefit of qemu support? If this is not absolutely needed, please remove it.

myungjin (commented):
left a few comments. Also, please add description in your commit with `git commit --amend`

.github/workflows/publish.yml
Divlik: This is to support multiple platform build, since we are not building for other platforms right now this can be safely removed.

main.go
Divlik: updated

myungjin (changes requested):
please update commit comments.

myungjin (approved):
lgtm

PR #270, opened 2022-11-18
Dataset Registration and Job Specification Changes

harshitdaga:
**Change Log**
Dataset registration: anyone should allow anyone to register their dataset. Once registered that dataset can be used by the developer submitting the job. This would help in decoupling the dataset with any machine learning job submission requests.

Job Specification: to provide improved structure create job specification would intake dataSpec and modelSpec details as separate files. Current PR provides the updated code for createJob, however this change has resulted in changes in earlier defined JobSpec object. Next PR would include the fix for other APIs such as updateJob, getJob etc.

**Suggestions** 
1. Remove re-defining errMsgFunc(...) example: pkg/openapi/controller/api_jobs_service.go. Instead abstracted it out into pkg/openapi/error.go as ErrorResponse(...)

2. pkg/openapi/controller/api_jobs_service.go createTasks(...) is calling database to obtain jobSpec. This function is called from 2 methods. One, when job spec is submitted and second, when job spec is updated. We should aim to reduce DB calls as much as possible and specifically in this scenario we can avoid it. Suggestion is to update the function and pass the job spec information from the method that is calling it.

**Ripple Effect Changes (changes to be handled during next PR)**
1. Files to be updated during TAG expansion fix
- cmd/controller/app/job/builder.go - setup() and buildTasks(...) 
- cmd/controller/app/objects/task.go - Configure(...) channel configuration
- pkg/openapi/controller/api_datasets_service.go Clear commented code related to createDataset


examples/local_testing/controller_config.yaml
```diff
@@ -0,0 +1,26 @@
+# Copyright 2022 Cisco Systems, Inc. and its affiliates
```
myungjin, line 1: I am not certain whether we need a local_testing folder. We can keep it for now for research_development branch. But eventually I feel that this shouldn't be included in the main repo. This is something to further discuss.

examples/local_testing/sample_job/job.json
```diff
@@ -0,0 +1,9 @@
+{
+    "designId": "test_design",
+    "schemaVersion": "1",
+    "codeVersion": "1",
+    "priority": "low",
+    "maxRunTime": 1800,
+    "dataSpecPath": "/Users/hd/Acads/projects/fledge/workspace/flame/examples/local_testing/sample_job/datas_spec.json",
+    "modelSpecPath": "/Users/hd/Acads/projects/fledge/workspace/flame/examples/local_testing/sample_job/datas_spec.json"
```
myungjin, line 8: These file path examples are incorrect. they won't work.
Also, please simply use a file name instead of a full path or relative path.
We can mandate all config json files are in the same folder.

When flamectl takes job.json as input, it can simply look for the path of job.json and try to access other json files from that directory. This simplifies the specification of these two fields.

myungjin (changes requested):
Thanks for the PR. This is a good start. Please consider to address comments.

One general comment is that this PR mixes two features (refactoring of job specification and redefinition of dataset) together. Consider to focus on one feature per a PR.

One way to quickly judge if a PR deals with more than one thing is to look at commit's title (or topic). If there is "and" in it, a PR tries to do more than one thing.

PR #273, opened 2022-11-21
chore: comparison of openapi directories

myungjin:
When openapi specification is updated, new files can be created by openapi.sh. One needs to know if file differences exist between flame/scripts/openapi and flame/pkg/openapi so that the differences can be reconciled judiciously.

A small script is created to compare the difference of files between the two directories.


Divlik (approved):
lgtm

PR #274, opened 2022-11-21
fix: mlflow flavor "keras" not found

myungjin:
In mlflow 2.0 or above, keras is not provided as a separate module. Instead, tensorflow module is used. When mlflow.keras is imported, mlflow.tensorflow is simply imported. So, trying to do getattr with string "keras" fails because there is no module called keras.

To fix the issue, when keras is identified, we use tensorflow flavor and call getattr with "tensorflow".


Divlik (approved):
lgtm

PR #275, opened 2022-11-22
Add --local-img flag to the shell script to use dev tag for local testing

Divlik:
Add args to start command for flame shell script to replace the tag name inside helmchart values file

### DIFF
--- a/fiab/flame.sh
+++ b/fiab/flame.sh
@@ -47,7 +47,13 @@ function pre_start_stop_check {
 
 function start {
     # install the flame helm chart
-    helm install --create-namespace --namespace $RELEASE_NAME $RELEASE_NAME helm-chart/control/
+    exposedb=$1
+    tag=$2
+    if [ "$tag" == "" ]; then
+        helm install --create-namespace --namespace $RELEASE_NAME $RELEASE_NAME helm-chart/control/
+    else
+        helm install --create-namespace --namespace $RELEASE_NAME $RELEASE_NAME helm-chart/control/ --set imageTag=$2
+    fi
 
     # Wait until mongodb is up
     ready=false
@@ -68,7 +74,7 @@ function start {
     done
     echo "done"
 
-    if [ "$1" == "false" ]; then
+    if [ "$exposedb" == "false" ]; then
         return
     fi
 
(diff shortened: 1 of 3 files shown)


fiab/flame.sh
```diff
@@ -206,16 +210,22 @@ function post_stop_cleanup {
 
 function main {
     exposedb=false
-    if [ "$2" == "--exposedb" ]; then
-        exposedb=true
-    fi
+    tag=""
+    for arg in "$@"; do
```
myungjin, line 214: This can be buggy because of line 218 (https://github.com/cisco-open/flame/pull/275/files#diff-c26314cc215f8cfbcb4f96f0ae8834a637f638dde7611c57148d786f9a3bfa48R218) which compares against a positional parameter.

I'd recommend the following command syntax:
./flame.sh <start [--exposedb] [--localimg] | stop>

I'd recommend to use "--localimg" to make the purpose of the argument is clearer. "--local" doesn't convey that we are using a local image.

The following example might be useful:
```
exposedb=false
tag=""
verb=${@:$OPTIND:1}
shift;

for arg in "$@"; do
    case "$arg" in
	"--exposedb") exposedb=true;;
	"--localimg") tag="dev";;
	*) echo "unknown option: $arg"; exit 1;;
    esac
done

echo $verb $exposedb $tag
```

Here you can use `verb` to replace `$1` in lines 223 and 233.

The usage syntax in line 243 should be updated (https://github.com/cisco-open/flame/pull/275/files#diff-c26314cc215f8cfbcb4f96f0ae8834a637f638dde7611c57148d786f9a3bfa48R243).

fiab/flame.sh
```diff
@@ -47,7 +47,11 @@ function pre_start_stop_check {
 
 function start {
     # install the flame helm chart
-    helm install --create-namespace --namespace $RELEASE_NAME $RELEASE_NAME helm-chart/control/
+    if [ "$2" == "" ]; then
```
myungjin, line 50: instead of using "$1" and "$2" in the start function, it may be better to assign variables for them.
```
exposedb=$1
tag=$2
```

myungjin (commented):
thanks for the PR. left two comments.

fiab/flame.sh
```diff
@@ -206,21 +212,31 @@ function post_stop_cleanup {
...
+
+    for arg in "$@"; do
+        case "$arg" in
+    	"--expose-db") exposedb=true;;
+    	"--local-img") tag="dev";;
+    	*) echo "unknown option: $arg"; exit 1;;
+        esac
+    done
+
+    echo $verb $exposedb $tag
```
myungjin, line 227: Remove this line. It's not necessary.

fiab/flame.sh
```diff
@@ -47,7 +47,13 @@ function pre_start_stop_check {
 
 function start {
     # install the flame helm chart
-    helm install --create-namespace --namespace $RELEASE_NAME $RELEASE_NAME helm-chart/control/
+    exposedb=$1
+    tag=$2
+    if [ "$tag" == "" ]; then
+      helm install --create-namespace --namespace $RELEASE_NAME $RELEASE_NAME helm-chart/control/
```
myungjin, line 53: Use four spaces for indentation.

myungjin (commented):
Minor comments left. Please address them. Otherwise, lgtm.

myungjin (approved):
lgtm

PR #277, opened 2022-11-23
Dataset Registration

harshitdaga:
Anyone should allow anyone to register their dataset. Once registered that dataset can be used by the developer submitting the job. This would help in decoupling the dataset with any machine learning job submission requests.


harshitdaga:
Goal of sample_experiment folder is to collect and provide all the configuration files needed by the developer if they like to run flame services locally and independently (not inside a docker container). Similar to fiab environment, I will write a small documentation on how that can be done.

api/dataset_api.partials.yml
```diff
@@ -226,4 +226,4 @@
         description: unexpected error
     summary: Update meta info for a given dataset
     tags:
-      - datasets
+      - datasets
```
myungjin, line 229: nit: restore a new line so that this unnecessary change is not in the PR.

examples/sample_experiment/controller_config.yaml
```diff
@@ -0,0 +1,26 @@
+# Copyright 2022 Cisco Systems, Inc. and its affiliates
```
myungjin, line 1: I believe that you are only adding this folder for development purpose only. I hope that this folder will be eventually removed from the examples folder because I don't feel it's an actual example. I even think its place is not correct. There should be a `test` folder and the sample_experiment folder should be there.

But I am not going to ask for any changes here as I believe that these should be eventually removed before this branch gets merged into the main.

pkg/openapi/controller/api_datasets_service.go
```diff
@@ -142,6 +110,48 @@ func (s *DatasetsApiService) UpdateDataset(ctx context.Context, user string, dat
 	return openapi.Response(http.StatusNotImplemented, nil), errors.New("UpdateDataset method not implemented")
 }
 
+/*
+TODO: Was used by the CreateDataset earlier function when we were passing computeID and Realm as part of the dataset.
+      As part of the new process dataset can be registered by anyone and the job creator will use datasetIDs during job submission.
+	  Code to be cleared after everything is merged and working.
+
+// CreateDataset - Create meta info for a new dataset.
+func (s *DatasetsApiService) CreateDataset(ctx context.Context, user string,
```
myungjin, line 119: Please consider to remove it if you don't need the commented-out part. Not sure about your intention here.

myungjin (approved):
left a few comments. there are minor points. so I approved it. I will leave a decision of reflecting my comments in this pr as yours.

examples/sample_experiment/controller_config.yaml
harshitdaga: Goal of sample_experiment folder is to collect and provide all the configuration files needed by the developer if they like to run flame services locally and independently (not inside a docker container). Similar to fiab environment, I will write a small documentation on how that can be done. 

We can place it under another folder/structure as your suggested but I felt this would be helpful for developers.

pkg/openapi/controller/api_datasets_service.go
harshitdaga: Intentionally left the comment with a TODO. Will be removed after TAG expansion changes are completed.

examples/sample_experiment/controller_config.yaml
myungjin: I understand your intention, but fiab is an env that developers can do development and testing "properly". Having a local setting is a great idea, but it comes with its own challenges and risks; for example, if you ran the filepath example in your previous PR in the fiab, you would have caught the issue. Also, developer needs additional facilities such as mongodb and mqtt broker (using a public mqtt broker is not ideal either as you don't have a "true" local setting).

If you really want this local setting, it needs to be done properly (e.g., by providing a script to configure everything completely in the host local system). If not done properly, developers may find it hard to use the local setting.

harshitdaga: Noted. Once I wrap my the TAG expansion I will address these concerns. If it makes sense we will keep it, otherwise I will add in a PR to remove these changes.

PR #281, opened 2022-11-30
Per Channel Backend

harshitdaga:
Now each channel can provide its communication backend protocol (optional). Additionally, default backend attribute (mandatory) is also added under job configuration. This information is used when per channel backend information is not provided. For example, if all the channels use the same backend or only a selective channel uses a different backend.


cmd/controller/app/objects/task.go
```diff
@@ -115,25 +112,3 @@ func (cfg *JobConfig) extractChannels(role string, channels []openapi.Channel) [
 
 	return exChannels
 }
-
-/*
```
myungjin, line 119: remove this instead of commenting it out.

examples/sample_experiment/job/job.json
```diff
@@ -0,0 +1,9 @@
+{
+    "designId": "test_design",
+    "schemaVersion": "1",
+    "codeVersion": "2",
+    "priority": "low",
+    "maxRunTime": 1800,
+    "dataSpecPath": "<file_path>/datas_spec.json",
+    "modelSpecPath": "<file_path>/model_spec.json"
```
myungjin, line 8: Let's enforce `dataSpecPath` and `modelSpecPath` files in the same directory where job.json is. Giving a full path is tedious and giving a relative path can be error-prone.

Also, this example is not correct.

Use the absolute path of https://github.com/cisco-open/flame/blob/research_development/cmd/flamectl/resources/job/job.go#L35 and search dataspec and modelspec files in that directory.

You can reflect this change in a separate PR if you want, but please consider to make that change.

myungjin (commented):
left two minor comments. Please consider to address them.

myungjin (commented):
left one comment.

myungjin (approved):
lgtm

PR #286, opened 2022-12-02
Deployer Registration Changes

harshitdaga:
Region is now called Realm. It is an array of tags used to describe geo-domain of the underlying cluster managed by the deployer. Additionally, a property attribute is added which will be used in the future to provide cluster properties such as GPU/TPU, storage information etc.. Such properties will be used by the ML developer to provide fine-grain specifications about the resources required to run their job.

### DIFF
--- a/api/computes_components.partials.yml
+++ b/api/computes_components.partials.yml
@@ -23,15 +23,25 @@ ComputeSpec:
   properties:
     adminId:
       type: string
-    region:
-      type: string
     apiKey:
       type: string
     computeId:
       type: string
+    realm:
+      type: array
+      items:
+        type: string
+    property:
+      type: object
+      additionalProperties:
+        type: string
+  required:
+    - adminId
+    - apiKey
+    - realm
   example:
     adminId: "admin-1"
-    region: "us-east"
+    realm: ["us-east", "us"]
     apiKey: "apikey-1"
     computeId: "compute-1"
 
--- a/cmd/deployer/cmd/root.go
+++ b/cmd/deployer/cmd/root.go
@@ -17,7 +17,9 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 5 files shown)


examples/sample_experiment/deployer_config.json
```diff
@@ -0,0 +1,11 @@
+{
```
myungjin, line 1: Eventually this needs to be included as part of fiab.

Look into some existing examples in the fiab: https://github.com/cisco-open/flame/blob/main/fiab/helm-chart/control/templates/controller-configmap.yaml and https://github.com/cisco-open/flame/blob/main/fiab/helm-chart/control/templates/controller-deployment.yaml#L46

api/computes_components.partials.yml
```diff
@@ -23,15 +23,23 @@ ComputeSpec:
...
     apiKey:
       type: string
     computeId:
       type: string
+    realm:
+      type: array
+      items:
+        type: string
+    property:
+      type: object
```
myungjin, line 35: What is the reason to make the value of property as `interface{}` (https://github.com/cisco-open/flame/pull/286/files#diff-0fffd693e21705492109d08469c06dce4672cdce49f455c8b9579e65fece7087R38)?
This basically suggests that the purpose of this attribute is "undefined/unknown".
I think we should start with a small scope (e.g., string) as value with the following syntax:
```
additionalProperties:
  type: string
```
Any opinion?

myungjin (commented):
left a couple of comments.

cmd/deployer/cmd/root.go
```diff
@@ -135,29 +120,8 @@ func init() {
...
-	rootCmd.MarkFlagRequired(argApiKey)
-
-	defaultPlatform := "k8s"
-	rootCmd.Flags().StringP(argPlatform, "p", defaultPlatform, "compute platform")
-	rootCmd.MarkFlagRequired(argPlatform)
-
-	defaultNamespace := "flame"
-	rootCmd.Flags().StringP(argNamespace, "s", defaultNamespace, "compute namespace")
-	rootCmd.MarkFlagRequired(argNamespace)
+	rootCmd.Flags().StringP(argConf, "c", "", "Deployer configuration file")
```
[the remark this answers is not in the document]
harshitdaga: P supports both single and longer format. User can use -c or --conf. My recommendation would be to keep both. But I have updated the code to use Var and only have longer format.

cmd/deployer/cmd/root.go
```diff
@@ -17,30 +17,32 @@
...
 
 const (
-	argApiserver = "apiserver"
-	argNotifier  = "notifier"
-	argAdminId   = "adminid"
-	argRegion    = "region"
-	argComputeId = "computeid"
-	argApiKey    = "apikey"
-	argPlatform  = "platform"
-	argNamespace = "namespace"
```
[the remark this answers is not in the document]
harshitdaga: Trying to avoid any technical debt :)

examples/sample_experiment/deployer_config.json
harshitdaga: Yes! I was going through the fiab codebase and saw these changes are required to be done later in fiab too. I will do that as part of future PR.

api/computes_components.partials.yml
harshitdaga: Good point! Spend some time going back and forth with your suggestion. However, decided to go with (type: object) because the additional properties can be bool (gpu support), number (max capacity) or string (storage capacity). Using string would make it restrictive.

Lmk what you think.

myungjin (approved):
lgtm

PR #295, opened 2022-12-12
updated fiab installation documents

RaviKhandavilli:
Made building flame image as an optional step on ubuntu and Macos documents


myungjin (approved):
lgtm

PR #296, opened 2022-12-13
updated deployer script to use public image

RaviKhandavilli:
Made changes in deployer.sh to use public flame image by default


myungjin (approved):
lgtm

PR #299, opened 2022-12-21
TAG Expansion Revamped

harshitdaga:
Updated the support for different topologies that can be compressed into TAG representation. The changes also include update in DataSpec representation and few testcases for different TAG topologies.


cmd/controller/app/objects/task.go
```diff
@@ -19,55 +19,69 @@ package objects
...
+	ModeSpec   openapi.ModelSpec `json:"modeSpec"`
+
+	//TODO maybe wrap it under DataSpec object?
+	DatasetUrl string `json:"dataset,omitempty"`
 }
 
 func (tsk *Task) generateTaskId(idx int) {
 	h := sha1.New()
-	data := fmt.Sprintf("%s-%d-%v", tsk.JobId, idx, tsk.JobConfig)
+	data := fmt.Sprintf("%s-%d-%v-%d", tsk.JobId, idx, tsk.JobConfig, time.Now().UnixNano())
```
myungjin, line 79: The idea of the original code is to ensure that the task ID is consistent based on some unique identifiers in the config. If time is involved, the task ID will be changed every time there is a small change in other tasks or other part of a schema.

Why is it necessary to make task ID different even if its content and identifiers didn't change?

cmd/controller/app/job/builder.go
```diff
@@ -17,11 +17,11 @@
 package job
```
myungjin, line 17: please run lint.sh and fix it if there is any issue.

myungjin (commented):
left two comments.

PR #319, opened 2023-02-01
Add group associations to roles

openwithcode:
## Description

With this PR we are adding group association for roles in order to define to which channel and group is associated with.

### DIFF
--- a/examples/medmnist/schema.json
+++ b/examples/medmnist/schema.json
@@ -2,34 +2,51 @@
     "name": "Benchmark of FedOPT Aggregators/Optimizers using MedMNIST example schema v1.0.0 via PyTorch",
     "description": "A simple example of MedMNIST using PyTorch to test out different aggregator algorithms.",
     "roles": [
-	{
-	    "name": "trainer",
-	    "description": "It consumes the data and trains local model",
-	    "isDataConsumer": true
-	},
-	{
-	    "name": "aggregator",
-	    "description": "It aggregates the updates from trainers"
-	}
+        {
+            "name": "trainer",
+            "description": "It consumes the data and trains local model",
+            "isDataConsumer": true,
+            "groupAssociation": [
+                {
+                    "param-channel": "default/us"
+                }
+            ]
+        },
+        {
+            "name": "aggregator",
+            "description": "It aggregates the updates from trainers",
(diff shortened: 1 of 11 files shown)


examples/medmnist/schema.json
```diff
@@ -2,34 +2,41 @@
...
-	    "isDataConsumer": true
-	},
-	{
-	    "name": "aggregator",
-	    "description": "It aggregates the updates from trainers"
-	}
+        {
+            "name": "trainer",
+            "description": "It consumes the data and trains local model",
+            "isDataConsumer": true
```
myungjin, line 8: perhaps `groupAssociation` field is missing.

examples/medmnist/schema.json
```diff
@@ -2,34 +2,41 @@
...
+        },
+        {
+            "name": "aggregator",
+            "description": "It aggregates the updates from trainers",
+            "replica": 1
```
myungjin, line 13: same here: `groupAssociation` field is missing.

pkg/openapi/model_role.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Cisco Systems, Inc. and its affiliates
+// Copyright 2023 Cisco Systems, Inc. and its affiliates
```
myungjin, line 1: The year should be the year that the work was introduced for the first time. So, please keep the year as is.

myungjin (commented):
left a couple of small comments. Please consider to address them.

examples/medmnist/schema.json
openwithcode: Indeed it was missing, i've added it here as well.

examples/medmnist/schema.json
openwithcode: Indeed it was missing, i've added it here as well.

pkg/openapi/model_role.go
openwithcode: You are right @myungjin the year should not change and should remain the year in which this file was created.

myungjin (commented):
lgtm. one last thing I'd like to ask is to squash all the commits into one so that we have a clean hash commit record.

myungjin (approved):

PR #322, opened 2023-02-03
documenation for metaserver

GustavBaumgart:
Documentation for using metaserver will allow users to run examples with a local broker. This decreases the chances of any job ID collisions.

Modifications to the config.json for the mnist example were made in order to make it easier to switch to a local broker. The readme does indicate how to do this for other examples now.

### DIFF
--- a/docs/08-flame-sdk.md
+++ b/docs/08-flame-sdk.md
@@ -20,17 +20,90 @@ make install
 ## Quickstart
 
 ### Configuring Brokers
-As the flame system uses MQTT brokers to exchange messages during federated learning, to run the python library locally, you could either 1) install a local MQTT broker 2) use a public MQTT broker. Here we'll illustrate the second option.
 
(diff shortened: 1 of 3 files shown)


docs/08-flame-sdk.md
```diff
@@ -20,9 +20,44 @@ make install
 ## Quickstart
 
 ### Configuring Brokers
-As the flame system uses MQTT brokers to exchange messages during federated learning, to run the python library locally, you could either 1) install a local MQTT broker 2) use a public MQTT broker. Here we'll illustrate the second option.
+Since the flame system uses MQTT brokers to exchange messages during federated learning, to run the python library locally, you could either 1) install a local MQTT broker 2) use a public MQTT broker.
```
myungjin, line 23: flame supports two backends: one is mqtt and the other is p2p.
For mqtt, we have two options, which are already stated.

The one you set up is for p2p backend. So, these need to be more clearly separated by leveraging different headings of markdown.

A broker for p2p backend is `metaserver` which we implement.
Please also add a note saying that `metaserver` is only for local testing; it is not for

myungjin (commented):
left a comment

myungjin (approved):
lgtm

PR #330, opened 2023-02-09
fix+refactor: asyncfl loss divergence

myungjin:
## Description

For asyncfl, a client (trainer) should send delta by subtracting local weights from original global weights after training. In the current implementation, the whole local weights were sent to a server (aggregator). This causes loss divergence.

Supporting delta update requires refactoring of aggregators of synchronous fl (horizontal/{top_aggregator.py, middle_aggregator.py}) as well as optimizers' do() function.

The changes here support delta update universally across all types of modes (horizontal synchronous, asynchronous, and hybrid).

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

### DIFF
--- a/lib/python/flame/common/typing.py
+++ b/lib/python/flame/common/typing.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 #
 # SPDX-License-Identifier: Apache-2.0
-
 """Definitions on Types."""
 
 from typing import Union
@@ -21,3 +20,6 @@ from typing import Union
 Scalar = Union[bool, bytes, float, int, str]
 
 Metrics = dict[str, Scalar]
+
+# list for tensorflow, dict for pytorach
+ModelWeights = Union[list, dict]
--- a/lib/python/flame/common/util.py
+++ b/lib/python/flame/common/util.py
@@ -21,11 +21,12 @@ import sys
 from contextlib import contextmanager
 from enum import Enum
 from threading import Thread
-from typing import List
+from typing import List, Union
 
 from pip._internal.cli.main import main as pipmain
 
 from ..config import Config
+from .typing import ModelWeights
 
 PYTORCH = 'torch'
 TENSORFLOW = 'tensorflow'
@@ -115,3 +116,21 @@ def mlflow_runname(config: Config) -> str:
                 groupby_value = groupby_value + val + "-"
 
     return config.role + '-' + groupby_value + config.task_id[:8]
+
+
+def delta_weights_pytorch(a: ModelWeights,
+                          b: ModelWeights) -> Union[ModelWeights, None]:
+    """Return delta weights for pytorch model weights."""
+    if a is None or b is None:
+        return None
+
+    return {x: a[x] - b[y] for (x, y) in zip(a, b)}
+
+
+def delta_weights_tensorflow(a: ModelWeights,
+                             b: ModelWeights) -> Union[ModelWeights, None]:
+    """Return delta weights for tensorflow model weights."""
+    if a is None or b is None:
+        return None
+
+    return [x - y for (x, y) in zip(a, b)]
--- a/lib/python/flame/examples/hier_mnist/middle_aggregator/config_uk.json
+++ b/lib/python/flame/examples/hier_mnist/middle_aggregator/config_uk.json
@@ -1,10 +1,14 @@
 {
     "taskid": "49d06b7526964db86cf37c70e8e0cdb6bd7aa743",
-    "backend": "mqtt",
+    "backend": "p2p",
     "brokers": [
         {
-            "host": "flame-mosquitto",
+            "host": "localhost",
             "sort": "mqtt"
+        },
+        {
+            "host": "localhost:10104",
+            "sort": "p2p"
         }
     ],
     "channels": [
@@ -64,12 +68,12 @@
 	"name": "mnist"
     },
     "registry": {
-	"sort": "mlflow",
+	"sort": "dummy",
 	"uri": "http://flame-mlflow:5000"
     },
     "selector": {
-	"sort": "random",
-	"kwargs": {"k": 1}
+	"sort": "default",
+	"kwargs": {}
     },
     "maxRunTime": 300,
     "realm": "default/uk/london/org2/flame",
--- a/lib/python/flame/mode/distributed/trainer.py
+++ b/lib/python/flame/mode/distributed/trainer.py
@@ -18,10 +18,12 @@
 # import hashlib
 import logging
 from collections import OrderedDict
+from copy import deepcopy
 
 from ...channel_manager import ChannelManager
 from ...common.custom_abcmeta import ABCMeta, abstract_attribute
-from ...common.util import (MLFramework, get_ml_framework_in_use,
+from ...common.util import (MLFramework, delta_weights_pytorch,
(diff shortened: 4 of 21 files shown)


lib/python/flame/common/typing.py
```diff
@@ -13,11 +13,12 @@
...
-
 """Definitions on Types."""
 
 from typing import Union
 
 Scalar = Union[bool, bytes, float, int, str]
 
 Metrics = dict[str, Scalar]
+
+ModelWeights = Union[list, dict]
```
elqurio, line 24: Not sure what to think about this. It seems wrong.
When I see ModelWeights, I assume it's a list, because of plurality. But it can be a list or a dict.
Maybe we need a ModelWeight type, and then if we want a list of these, we can do list[ModelWeight]

lib/python/flame/common/util.py
```diff
@@ -115,3 +116,21 @@ def mlflow_runname(config: Config) -> str:
...
+    return {x: a[x] - b[y] for (x, y) in zip(a, b)}
+
+
+def delta_weights_tensorflow(a: ModelWeights,
+                             b: ModelWeights) -> Union[ModelWeights, None]:
+    """Return delta weights for tensorflow model weights."""
+    if a is None or b is None:
+        return None
+
+    return [x - y for (x, y) in zip(a, b)]
```
elqurio, line 136: Ah, so for pytorch, the weights are dict, but for tensorflow they are a list.
Seems to me like we can implement a factory here. But the typing now makes more sense.

lib/python/flame/examples/hier_mnist/middle_aggregator/config_uk.json
```diff
@@ -1,10 +1,14 @@
 {
     "taskid": "49d06b7526964db86cf37c70e8e0cdb6bd7aa743",
-    "backend": "mqtt",
+    "backend": "p2p",
```
elqurio, line 3: Are these changes here by mistake? Don't see the connection with this PR

lib/python/flame/mode/distributed/trainer.py
```diff
@@ -53,16 +55,14 @@ def init_cm(self) -> None:
...
         self.registry_client = registry_provider.get(self.config.registry.sort)
         # initialize registry client
         self.registry_client(self.config.registry.uri, self.config.job.job_id)
 
         base_model = self.config.base_model
         if base_model and base_model.name != "" and base_model.version > 0:
             self.model = self.registry_client.load_model(
-                base_model.name, base_model.version
-            )
+                base_model.name, base_model.version)
```
elqurio, line 65: We should consider adding a auto formatter, like black to prevent these dust changes

lib/python/flame/mode/horizontal/asyncfl/middle_aggregator.py
```diff
@@ -161,6 +176,9 @@ def _aggregate_weights(self, tag: str) -> None:
             time.sleep(1)
             return
 
+        # save global weights before updating it
+        self.prev_weights = self.weights
```
elqurio, line 180: Should this be a deepcopy as well?

lib/python/flame/mode/horizontal/trainer.py
```diff
@@ -143,6 +153,9 @@ def _update_model(self):
             self.model.set_weights(self.weights)
 
     def _update_weights(self):
+        # save weights before updating it
+        self.prev_weights = self.weights
```
elqurio, line 157: Deepcopy?

elqurio (commented):
Apart from config changes in examples, the branching for PyTorch and TF, and a couple of deepcopies, it looks good to me.
As for examples, shouldn't we add a different example for asyncFL rather than updating existing ones which are synchronous?

lib/python/flame/common/typing.py
myungjin: This can be confusing. A ML model can be represented as list or dict depending on the framework (pytorch or tensorflow). Each is ModelWeights. The reason I use the plural form rather than a singular form is that weight usually means one value in the list or dict.

lib/python/flame/common/util.py
myungjin: a factory pattern can be implemented with design changes, which require more thinking.

lib/python/flame/examples/hier_mnist/middle_aggregator/config_uk.json
myungjin: you're right. They aren't.
Now that we have local setting doc (https://github.com/cisco-open/flame/blob/main/docs/08-flame-sdk.md#quickstart), I wanted to align the examples with the doc.

Instead of creating a separate PR, I piggybacked these into this PR. If you want to separate a PR, I can remove them from here. let me know.

lib/python/flame/mode/distributed/trainer.py
myungjin: I am using python lsp. sometimes, it goes back and forth with different format.
If we can enforce auto-formatting before PR submission, that would be nice. But I don't know how.

lib/python/flame/mode/horizontal/asyncfl/middle_aggregator.py
myungjin: No, it's not necessary because I already did deepcopy at https://github.com/cisco-open/flame/pull/330/files#diff-074d31d6330e5c3880f78eef8d7bfbcfd61746fe1de6a990d07cbc48c61ebcafR134.

To summarize the structure,

self._agg_goal_weight = deepcopy(self.weights) // let's say the address of self._agg_goal_weights is `abc` and the address of self.weights is `def`
// do some aggregation with self._agg_goal_weight.

// once some condition is met, we reach the following statement
self.prev_weights = self.weights // addr(prev_weights) = `def`
self.weights = self._agg_goal_weights // addr(weights) = `abc`

lib/python/flame/mode/horizontal/trainer.py
myungjin: we are safe here too because of https://github.com/cisco-open/flame/pull/330/files#diff-148a2e70215f277621faa325f4be3ab1be7ca28f9729e3732e7ea86c89eca326R97.

self.weights = self.weights = msg[MessageType.WEIGHTS] // let's say addr(weights) = `abc`.
then we call `self._update_model()`, which load weights into a model (e.g., by doing `self.model.load_state_dict(self.weights)` in `self._update_model()`). This won't let weights in model point to `abc`.

So, when we call `self.prev_weights = self.weights` , addr(self.prev_weights) = `abc`. when we call `self.weights = self.model.state_dict()` for example, addr(self.weights) may be some other value than `abc`.

Also, prev_weights and weights point to the same object, delta_weights would have been zero.

lib/python/flame/common/util.py
elqurio: Agreed, it's a much bigger change than this PR.

lib/python/flame/mode/distributed/trainer.py
elqurio: I can set up pre commit hooks to run black I think. Also one of the tools I like is isort, it sorts imports in a specific order (stdlib > installed libs > project imports). And it has a profile to work well with black, because black likes to do some formatting on sorts as well.

PR #332, opened 2023-02-13
Configuration Handling Refactor

elqurio:
## Description

Due to the current structure and the difference between the config that is provided, and the config that is actually consumed (and how it's consumed), I had to create transformers for the json data. Their purpose is to transform the provided json config and turn it into a structure that pydantic models can consume without issues.

I think our goal should be to keep the structure provided as close as possible to the structure that is consumed, just to make things more straightforward. I'd like to believe that we can completely remove transformers, and consume data in the same structure it is provided.

This PR should bring more clarity into what is really happening during the configuration loading, making it much easier to change and further improve.

Also, typing support is never a bad thing, plus we can export the json schema of the models which is an additional benefit.
## Type of Change

- [ ] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/lib/python/flame/config.py
+++ b/lib/python/flame/config.py
@@ -15,419 +15,268 @@
 # SPDX-License-Identifier: Apache-2.0
 """Config parser."""
 
-import json
-import sys
 from enum import Enum
-
-CONF_KEY_TASK = 'task'
-CONF_KEY_TASK_ID = 'taskid'
-CONF_KEY_BACKEND = 'backend'
-CONF_KEY_CHANNEL = 'channels'
-CONF_KEY_DATASET = 'dataset'
-CONF_KEY_HYPERPARAMS = 'hyperparameters'
-CONF_KEY_MAX_RUN_TIME = 'maxRunTime'
-CONF_KEY_REALM = 'realm'
-CONF_KEY_ROLE = 'role'
-
-CONF_KEY_CHANNEL_NAME = 'name'
-CONF_KEY_CHANNEL_PAIR = 'pair'
-CONF_KEY_CHANNEL_IS_BIDIR = 'isBidirectional'
-CONF_KEY_CHANNEL_GROUPBY = 'groupBy'
-CONF_KEY_CHANNEL_GROUPBY_TYPE = 'type'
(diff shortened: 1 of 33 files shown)


lib/python/flame/config.py
```diff
@@ -1,433 +1,258 @@
-# Copyright 2022 Cisco Systems, Inc. and its affiliates
```
myungjin, line 1: Please keep the copyright.

lib/python/flame/config.py
```diff
@@ -1,433 +1,258 @@
...
+    dataset: str
+    max_run_time: int
+    base_model: BaseModel
+    groups: t.Optional[Groups]
+    dependencies: list[str]
+    func_tag_map: t.Optional[dict]
+
+
+def read_config() -> dict:
+    with open("flame/examples/mnist/trainer/config.json") as f:
```
myungjin, line 152: Why is a filename hard-coded? Is this for testing purpose? If this is intended, please leave a comment. If not, please add `TODO`.

myungjin (commented):
left two comments. please consider to address them.

lib/python/flame/config.py
```diff
@@ -1,433 +1,258 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-# SPDX-License-Identifier: Apache-2.0
-"""Config parser."""
-
-import json
-import sys
 from enum import Enum
+from pydantic import Field
```
myungjin, line 2: If `pydantic` is not one of default python packages, please update dependencies in `setup.py` (i.e., `install_requires`).

lib/python/flame/config.py
elqurio: Fixed this, it accepts an argument filename which is passed during initialisation.

lib/python/flame/config.py
elqurio: Added, nice catch!

myungjin (approved):
lgtm

PR #337, opened 2023-02-15
Add per channel backend configuration

openwithcode:
## Description

Until now we were configuring the job with the required backend protocol used to communicate.
Now we allow each channel to define it's own backend communication protocol. 
If no communication protocol it's specified then it will be used the one witch is default. The default communication protocol it's a new field defined at schema level and which it's mandatory to be specified.

## Type of Change

- [ ] Bug Fix
- [ ] New Feature
- [x] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have read the [contributing guidelines](/CONTRIBUTING.md)
- [x] I have verified this change is not present in other open pull requests

### DIFF
--- a/lib/python/flame/examples/adult/aggregator/config.json
+++ b/lib/python/flame/examples/adult/aggregator/config.json
@@ -1,15 +1,16 @@
 {
     "taskid": "49d06b7526964db86cf37c70e8e0cdb6bd7aa742",
-    "backend": "mqtt",
     "brokers": [
         {
             "host": "flame-mosquitto",
             "sort": "mqtt"
         }
     ],
+    "defaultBackend": "mqtt",
     "channels": [
         {
             "description": "Model update is sent from trainer to aggregator and vice-versa",
+            "backend": "mqtt",
             "groupBy": {
                 "type": "tag",
                 "value": [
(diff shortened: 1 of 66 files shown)


lib/python/flame/examples/adult/aggregator/config.json
```diff
@@ -1,15 +1,16 @@
 {
     "taskid": "49d06b7526964db86cf37c70e8e0cdb6bd7aa742",
-    "backend": "mqtt",
     "brokers": [
         {
             "host": "flame-mosquitto",
             "sort": "mqtt"
         }
     ],
+    "defaultBackend": "mqtt",
```
myungjin, line 9: This change requires the support from config.py. Can you please sync up with @elqurio and make sure that you two are on the same page unless you have already done so?

lib/python/flame/examples/adult/aggregator/config.json
```diff
@@ -1,15 +1,16 @@
...
     "channels": [
         {
             "description": "Model update is sent from trainer to aggregator and vice-versa",
+            "backend": "mqtt",
```
myungjin, line 13: same here! please connect with @elqurio.

@elqurio this part's functionality also overlaps with `channelConfigs` (https://github.com/cisco-open/flame/blob/main/lib/python/flame/examples/hybrid/trainer/config_eu_org1.json#L84).

The key difference is that this part doesn't allow to choose different endpoint for the same backend whereas the channelConfigs has `brokers` block which allows user can use different `host`.

I feel that this needs some thought. Can you folks sync up with each other?

myungjin (commented):
left a few comments. please make sure that you sync up with @elqurio .

PR #338, opened 2023-02-16
Create a script to run minikube

alexandruuBytex:
## Description

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/scripts/runMinikube.sh
+++ b/scripts/runMinikube.sh
@@ -1,3 +1,55 @@
+#!/usr/bin/env bash
+# Copyright 2023 Cisco Systems, Inc. and its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+echo "
+########## Attention! ##########
+# This is a MacOS only, internal utility!
+# This is supposed to be used only for the development of Flame, only on MacOS devices!
+# This script will stop and recreate your minikube environment.
+# Run this script at your own risk!
+################################
+"
+
+if [[ $(uname) != "Darwin" ]]; then
+    echo "This script is supposed to run only on MacOS devices!"
+    exit 1
+fi
+
+force=false
+
+for arg in "$@"; do
+    case "$arg" in
+        "-f")
+            force=true
+            ;;
+        *)
+            echo "unknown option: $arg";exit 1
+            ;;
+    esac
+done
+
+if [[ $force == false ]]; then
+    read -p "Are you sure you want to continue? y or n: " </dev/tty response
+
+    if [[ $response == 'n' ]]; then
+        exit 0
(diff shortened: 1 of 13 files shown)


scripts/runMinikube.sh
```diff
@@ -0,0 +1,32 @@
...
+minikube start --cpus=6 --memory=6g --disk-size 100gb
+
+echo "Running: minikube ssh \"sudo systemctl stop systemd-resolved\""
+minikube ssh "sudo systemctl stop systemd-resolved"
+
+echo "Running: minikube ssh \"sudo systemctl disable systemd-resolved\""
+minikube ssh "sudo systemctl disable systemd-resolved"
+
+echo "Running: minikube ssh \"printf 'nameserver 8.8.8.8\nsearch .\n' | sudo tee /etc/resolv.conf\""
+minikube ssh "printf 'nameserver 8.8.8.8\nsearch .\n' | sudo tee /etc/resolv.conf"
```
myungjin, line 17: Lines 11-17 are only needed for a corner case where dnscrypt-proxy or dnsmasq is running. So, blindly running these commands is not ideal.

Also, line 8 is also hard-coded with cpu, memory and disk-size.

This may not be for general use by users. If it is only for internal use, we need to let users know the ramification of using this script.

At a minimum, this script should print out some informative messages (and warning) and receive user's consent if proceeding with this script is okay.

scripts/runMinikube.sh
```diff
@@ -0,0 +1,32 @@
+echo "Running: minikube stop"
+minikube stop
+
+echo "Running: minikube delete"
+minikube delete
```
myungjin, line 5: Please remove `minikube delete`. Let user explicitly delete minikube before he/she runs this script.
It's not obvious that running this script means deletion of minikube by just looking at the name of the script.

myungjin (commented):
This script can be very useful. Thanks for this. However, I also have some concern too; the script is too static (everything is hard-coded).

Also, I have impression that this is only for internal use. If not, the script should be under fiab folder and also needs a proper documentation. Assuming that this is only for internal use, we might want to minimize a chance that users may run this script accidentally and run into unexpected situations (e.g., deleting properly configured minikube env).

In the long run, this should be under fiab folder.

scripts/runMinikube.sh
```diff
@@ -0,0 +1,26 @@
+set -x;
+
+minikube stop
+
+minikube start --driver=hyperkit --cpus=6 --memory=6g --disk-size 100gb --kubernetes-version=v1.23.8 --alsologtostderr -v=7
+
+minikube ssh "echo 'cat << EOF >  /var/lib/boot2docker/bootlocal.sh
```
myungjin, line 7: lines 7-14 are for a case where dnscrypt-proxy or dnsmasq is running.
Please take time to consider my earlier comment: https://github.com/cisco-open/flame/pull/338#discussion_r1108912715.

Again, if this script is only for internal developers, please put some guard against misuse.

scripts/runMinikube.sh
```diff
@@ -0,0 +1,26 @@
...
+
+../fiab/setup-cert-manager.sh
+
+eval $(minikube docker-env)
+
+../fiab/build-image.sh 
+
+afplay /System/Library/Sounds/Glass.aiff
+
+set +x;
```
myungjin, line 26: nit: add new line

myungjin (commented):
left a couple of comments.

scripts/runMinikube.sh
```diff
@@ -0,0 +1,26 @@
+set -x;
```
myungjin, line 1: add a license header

PR #339, opened 2023-02-16
example/implementation for fedprox

GustavBaumgart:
This example is similar to the ones seen in the fedprox paper, although it currently does not simulate stragglers and uses another dataset/architecture.

A few things were changed in order for there to be a simple process for modifying trainers. This includes a function in util.py and another class variable in the trainer containing information on the client side regularizer.

Additionally, tests are automated (mu=1,0.1,0.01,0.001,0) so running the example generates or modifies existing files in order to provide the proper configuration for an experiment.


lib/python/flame/optimizer/fedprox.py
```diff
@@ -0,0 +1,46 @@
...
+        self.mu = mu
+        ml_framework_in_use = get_ml_framework_in_use()
+
+        # only support pytorch for fedprox
+        if ml_framework_in_use != MLFramework.PYTORCH:
+            raise NotImplementedError(
+                "supported ml framework not found; "
+                f"supported frameworks (for fedprox) are: {[MLFramework.PYTORCH.name.lower()]}")
+        
+        super().__init__()
```
myungjin, line 42: would it be safe to call this before initializing the child class's member variables?

lib/python/flame/mode/horizontal/trainer.py
```diff
@@ -59,6 +60,11 @@ def internal_init(self) -> None:
         self.registry_client.setup_run(mlflow_runname(self.config))
         self.metrics = dict()
 
+        # needed in fedprox trainer
```
myungjin, line 63:
```suggestion
        # needed for trainer-side optimization algorithms such as fedprox
```

myungjin (commented):
left several comments; please consider to address them.

lib/python/flame/optimizer/fedavg.py
```diff
@@ -22,6 +22,7 @@
 from ..common.util import (MLFramework, get_ml_framework_in_use,
                            valid_frameworks)
 from .abstract import AbstractOptimizer
+from .regularizer.regularizer import Regularizer
```
myungjin, line 25: This naming convention (`regularizer.regularizer`) looks a bit weird.
It may be better to rename regularizer.py to default.py, dummy.py or something else other than regularizer.

myungjin (commented):
left one comment

myungjin (approved):
lgtm

PR #350, opened 2023-02-28
gpu and cpu compatibility pytorch

GustavBaumgart:
This modification will allow trainers/aggregators to work on different devices across different machines.

All weights are communicated by placing them on the CPU before serializing them. Then, they are moved back to the device where they were previously.

An import error for fedprox was also fixed.
All trainers now require a self.model attribute, and all middle aggregators must only use the CPU (not enforced).

### DIFF
--- a/lib/python/flame/common/util.py
+++ b/lib/python/flame/common/util.py
@@ -27,6 +27,7 @@ from pip._internal.cli.main import main as pipmain
 
 from ..config import Config
 from .typing import ModelWeights
+from .constants import DeviceType
 
 PYTORCH = 'torch'
 TENSORFLOW = 'tensorflow'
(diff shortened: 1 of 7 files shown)


lib/python/flame/common/util.py
```diff
@@ -137,3 +138,32 @@ def delta_weights_tensorflow(a: ModelWeights,
...
+
+def get_cpu_device_pytorch():
+    import torch
+    return torch.device('cpu')
+
+def get_gpu_device_pytorch():
+    import torch
+    return torch.device('cuda')
+
+def weights_to_device(weights, device):
```
elqurio, line 150: `weights_to_device` doesn't really explain what is happening.  The functions seems to convert/translate weights. I think `covert_weights_for_dtype` is a more descriptive name. And the argument `device` isn't really a device, but its type, which is deceiving - either a type hint or a name change (or both) would be nice.

If you combine the above implementation, it should allow a refactor along the lines of:
```pseudo
elif pytorch:
    torch_device = get_pytorch_device(dtype)
    return {name: weights[name].to(torch_device) for name in weights}
else:
    raise ? # not sure what needs to happen here, I guess raise and exit, because it would mean we couldn't detect the framework, or the framework we detected is not tf nor pytorch
```

lib/python/flame/common/util.py
```diff
@@ -137,3 +138,32 @@ def delta_weights_tensorflow(a: ModelWeights,
...
+        if device == DeviceType.CPU:
+            torch_device = get_cpu_device_pytorch()
+            return {name: weights[name].to(torch_device) for name in weights}
+        elif device == DeviceType.GPU:
+            torch_device = get_gpu_device_pytorch()
+            return {name: weights[name].to(torch_device) for name in weights}
+
+    return None
+
+def weights_to_model_device(weights, model):
```
elqurio, line 163: I don't quite understand what this does. What is the difference between the `weights_to_device` and `weights_to_model_device` conceptually?

elqurio (changes requested):
Looks good overall, couple of notes in the comments.
Does this PR impact the examples? Where will the `trainer.model` come from?
Also, we should probably find a way to tell what framework we are using without having to check `sys.modules` every time (not in this PR though).

elqurio (approved):

PR #354, opened 2023-03-02
fix: out of order message delivery

myungjin:
## Description

In order to minimize the blocking of asynchronous I/O tasks when a large number of message chunks (e.g., model size is large) need to be assembled, threading is used. This led to a bug wherein messages are delivered to a receive queue in an out-of-order fashion. For example, consider two back-to-back messages; the first message is larger than the second. In this case, the second can be inserted before the first because the second has high chance to finish the assembling.

The changes here are for fixing the bug. To do so, we leverage another queue to synchronize message delivery order. For that, chunk manager class is introduced. One chunk manager is created for a backend. And it maintains a list of chunk threads per end, each of which is responsible for assemling messages sequentially and pushing them into the receive queue of the end.

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/lib/python/flame/backend/chunk_manager.py
+++ b/lib/python/flame/backend/chunk_manager.py
@@ -1,0 +1,115 @@
+# Copyright 2023 Cisco Systems, Inc. and its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


elqurio (approved):
Some of the commends are self explanatory. But the approach seems valid

PR #365, opened 2023-03-09
add download file function

GustavBaumgart:
The download function can take in the config's dataset field and return the location of the file where its stored (or downloaded, if that's needed). This makes it easier to load files in aggregator/main.py and trainer/main.py.

### DIFF
--- a/lib/python/flame/common/util.py
+++ b/lib/python/flame/common/util.py
@@ -77,6 +77,27 @@ def get_params_detached_pytorch(model):
     """Return copy of parameters of pytorch model disconnected from graph."""
     return [param.detach().clone() for param in model.parameters()]
 
+def get_dataset_filename(link):
+    """Return path for file location"""
+    # currently only supports https and local file
+    if link.startswith('https://'):
+        import requests
+        r = requests.get(link, allow_redirects=True)
+
+        try:
+            filename = link.split('/')[-1]
+            open(filename, 'wb').write(r.content)
+        except:
+            filename = 'data'
+            open(filename, 'wb').write(r.content)
+            
+        return filename
+        
+    elif link.startswith('file://'):
+        return link[7:]
+
+    raise TypeError('link format not supported; use either https:// or file://')
+
 @contextmanager
 def background_thread_loop():
 


lib/python/flame/common/util.py
```diff
@@ -77,6 +77,27 @@ def get_params_detached_pytorch(model):
...
+        except:
+            filename = 'data'
+            open(filename, 'wb').write(r.content)
+            
+        return filename
+        
+    elif link.startswith('file:'):
+        return link[5:]
+
+    return None
```
elqurio, line 99: Download would do nothing if link doesn't start with either file: or https://, do we want to raise instead maybe?

lib/python/flame/common/util.py
```diff
@@ -77,6 +77,27 @@ def get_params_detached_pytorch(model):
...
+        r = requests.get(link, allow_redirects=True)
+
+        try:
+            filename = link.split('/')[-1]
+            open(filename, 'wb').write(r.content)
+        except:
+            filename = 'data'
+            open(filename, 'wb').write(r.content)
+            
+        return filename
```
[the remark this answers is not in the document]
GustavBaumgart: Yeah, I don't know what the extension would be then, so I just gave a default file name in the case that there is a like an illegal character in the link.

lib/python/flame/common/util.py
GustavBaumgart: Maybe that's a better practice, yes. Do you know what exception to use or is it just a custom made one?

lib/python/flame/common/util.py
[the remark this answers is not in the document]
elqurio: What I was going for is, `file:` URL will contain the full path, since you return link[5:] in that case, the filename will be the entire filepath.
However in the case for `https://` the returned filename will be only the filename, not filepath, right?

lib/python/flame/common/util.py
elqurio: Maybe TypeError can work with an appropriate message

lib/python/flame/common/util.py
```diff
@@ -77,6 +77,27 @@ def get_params_detached_pytorch(model):
...
+        
+    elif link.startswith('file:'):
```
myungjin, line 96: it should be "file://" not "file:".

lib/python/flame/common/util.py
```diff
@@ -77,6 +77,27 @@ def get_params_detached_pytorch(model):
     """Return copy of parameters of pytorch model disconnected from graph."""
     return [param.detach().clone() for param in model.parameters()]
 
+def download(link):
```
myungjin, line 80: Use this method in the examples so that you can check if this method works or not.

lib/python/flame/common/util.py
```diff
@@ -77,6 +77,27 @@ def get_params_detached_pytorch(model):
...
+            filename = link.split('/')[-1]
+            open(filename, 'wb').write(r.content)
+        except:
+            filename = 'data'
+            open(filename, 'wb').write(r.content)
+            
+        return filename
+        
+    elif link.startswith('file:'):
+        return link[5:]
```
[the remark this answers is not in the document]
GustavBaumgart: Not completely sure what you mean by this

lib/python/flame/common/util.py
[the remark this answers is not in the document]
GustavBaumgart: This is correct. I would just use the current working directory when in doubt.

elqurio (approved):
LGTM

PR #373, opened 2023-03-21
feat: Add per-channel backend support

alexandruuBytex:
## Description

If the job backend is not specified and at least one channel does not define any backend, one backend type from the brokers will be used.

## Type of Change

- [ ] Bug Fix
- [ x ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


cmd/controller/app/job/builder.go
```diff
@@ -189,6 +189,58 @@ func (b *JobBuilder) setup() error {
 }
 
 func (b *JobBuilder) build() ([]objects.Task, []string, error) {
+	var channels map[string]openapi.Channel = b.channels
+	var brokers []config.Broker = b.jobParams.Brokers
+
+	var jobBackend string = string(b.jobSpec.Backend)
```
openwithcode, line 195: We should not cast it to string, but we should compare if the value contained in the b.jobSpec.Backend is one of the valid types

cmd/controller/app/job/builder.go
```diff
@@ -189,6 +189,58 @@ func (b *JobBuilder) setup() error {
...
+	var brokerBackend []string
+	var incompleteChannels []openapi.Channel
+
+	var validBackends = []string{"mqtt", "p2p"}
+	var validBackendFound = false
+
+	for _, channel := range channels {
+		incompleteChannels = append(incompleteChannels, channel)
```
openwithcode, line 203: Is this really useful ?

cmd/controller/app/job/builder.go
```diff
@@ -189,6 +189,58 @@ func (b *JobBuilder) setup() error {
 }
 
 func (b *JobBuilder) build() ([]objects.Task, []string, error) {
```
openwithcode, line 191: Please export the changes to a separate function that cares of per channel checks

cmd/controller/app/job/builder.go
```diff
@@ -189,6 +189,58 @@ func (b *JobBuilder) setup() error {
...
+
+	for _, channel := range channels {
+		incompleteChannels = append(incompleteChannels, channel)
+	}
+
+	if jobBackend == "" && len(incompleteChannels) > 0 {
+		return nil, nil, fmt.Errorf("Validation failed! Job backend is required!")
+	}
+
+	if len(brokers) < len(incompleteChannels) {
```
openwithcode, line 210: This is not really correct as we can have a single broker for many channels

cmd/controller/app/job/builder.go
```diff
@@ -189,6 +189,58 @@ func (b *JobBuilder) setup() error {
...
+	var incompleteChannels []openapi.Channel
+
+	var validBackends = []string{"mqtt", "p2p"}
+	var validBackendFound = false
+
+	for _, channel := range channels {
+		incompleteChannels = append(incompleteChannels, channel)
+	}
+
+	if jobBackend == "" && len(incompleteChannels) > 0 {
```
openwithcode, line 206: backend is optional.
If is not specified then we need to choose a random one from the broker list that we have. This is applied only if the user doesn't specify the backend for at least one of the channels

cmd/controller/app/job/builder.go
```diff
@@ -189,6 +189,58 @@ func (b *JobBuilder) setup() error {
...
+	}
+
+	if len(brokers) < len(incompleteChannels) {
+		return nil, nil, fmt.Errorf("Validation failed! There are not enough brokers for the incomplete channels!")
+	}
+
+	for _, validBackend := range validBackends {
+		for _, broker := range b.jobParams.Brokers {
+			if broker.Sort == validBackend {
+				validBackendFound = true
```
openwithcode, line 217: the for loop should be stopped after the backend it's matched

cmd/controller/app/job/builder.go
```diff
@@ -189,6 +189,58 @@ func (b *JobBuilder) setup() error {
...
+			}
+		}
+
+		if jobBackend == validBackend {
+			validBackendFound = true
```
openwithcode, line 222: same as above

cmd/controller/app/job/builder.go
```diff
@@ -189,6 +189,58 @@ func (b *JobBuilder) setup() error {
...
+
+	if !validBackendFound {
+		return nil, nil, fmt.Errorf("Validation failed! Invalid backend inside the broker!")
+	}
+
+	if len(brokerBackend) == 1 && brokerBackend[0] != jobBackend {
+		return nil, nil, fmt.Errorf("Validation failed!")
+	}
+
+	for index, channel := range incompleteChannels {
```
openwithcode, line 234: I don't think that we should populate the backend and host for every channel as it's not really applied based on the design schema we have.
We have to add it from schema.json to the generated config file if is specified

openwithcode (changes requested):

pkg/openapi/model_channel.go
```diff
@@ -29,6 +29,10 @@ package openapi
 type Channel struct {
 	Name string `json:"name"`
 
+	Backend string `json:"backend"`
+
+	Host string `json:"host"`
```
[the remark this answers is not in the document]
alexandruuBytex: Should I change host into broker host in the json file too?

pkg/openapi/model_channel.go
```diff
@@ -29,6 +29,10 @@ package openapi
 type Channel struct {
 	Name string `json:"name"`
 
+	Backend CommBackend `json:"backend"`
+
+	BrokerHost string `json:"host"`
```
openwithcode, line 34: the json tag `host` should be renamed as well

cmd/controller/app/job/builder.go
```diff
@@ -188,7 +188,57 @@ func (b *JobBuilder) setup() error {
...
+
+	if len(brokers) == 1 && openapi.CommBackend(brokers[0].Sort) != jobBackend {
+		return fmt.Errorf("Validation failed!")
+	}
+
+	return nil
+}
+
 func (b *JobBuilder) build() ([]objects.Task, []string, error) {
+	if err := b.perChannelBackendCheck(); err != nil {
```
openwithcode, line 238: Maybe we can add this directly in the setup method

examples/mnist/dataSpec.json
```diff
@@ -1,12 +1,12 @@
 {
-    "fromUser": {
```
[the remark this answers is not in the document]
alexandruuBytex: I installed it and updated the file. I was using a tab size of 4 spaces in the vs code settings but another setting was overwriting the first one.
Now it seems to work properly.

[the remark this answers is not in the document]
openwithcode: The formatting is still applied to this file. Please revert the change

examples/mnist/job.json
```diff
@@ -1,11 +1,11 @@
 {
-    "designId": "mnist",
```
openwithcode, line 2: please revert  the formatting change here as weell

examples/mnist/schema.json
```diff
@@ -1,52 +1,52 @@
 {
-    "name": "A simple example schema v1.0.0",
```
openwithcode, line 2: revert the formatting changes of this file

openwithcode (changes requested):

cmd/controller/app/job/builder_example_test.go
```diff
@@ -60,7 +60,18 @@ func Test_asyncfl_hier_mnist(t *testing.T) {
 	dbService, err := mongodb.NewMongoServiceWithClient(ctx, db)
 	assert.NoError(t, err)
 
-	builder := NewJobBuilder(dbService, config.JobParams{})
+	builder := NewJobBuilder(dbService, config.JobParams{
```
openwithcode, line 63: export the brokers and just reference them as at this point it's a lot of duplicated code

cmd/controller/app/job/builder_example_test.go
```diff
@@ -202,7 +224,18 @@ func Test_hier_mnist(t *testing.T) {
...
+		Brokers: []config.Broker{
```
openwithcode, line 228: same as above

cmd/controller/app/job/builder_example_test.go
```diff
@@ -276,7 +309,18 @@ func Test_medmnist(t *testing.T) {
...
```
openwithcode, line 313: same as above

cmd/controller/app/job/builder_example_test.go
```diff
@@ -371,7 +415,18 @@ func Test_mnist(t *testing.T) {
...
```
openwithcode, line 419: same as above

cmd/controller/app/job/builder_example_test.go
```diff
@@ -434,7 +489,18 @@ func Test_parallel_experiment(t *testing.T) {
...
```
openwithcode, line 493: same as above

cmd/controller/app/job/testdata/asyncfl_hier_mnist/1.json
```diff
@@ -10,6 +10,16 @@
     },
     "role": "middle-aggregator",
     "realm": "",
+	"brokers": [
```
openwithcode, line 13: formatting is wrong

cmd/controller/app/job/testdata/asyncfl_hier_mnist/1.json
```diff
@@ -10,6 +10,16 @@
...
+		{
```
openwithcode, line 14: given the fact that no channel specified any backend, then don't add p2p as it will not be used

cmd/controller/app/job/testdata/asyncfl_hier_mnist/2.json
```diff
@@ -10,6 +10,16 @@
     },
     "role": "middle-aggregator",
     "realm": "",
+	"brokers": [
```
openwithcode, line 13: same as above

openwithcode (changes requested):

pkg/openapi/model_channel.go
```diff
@@ -29,6 +29,10 @@ package openapi
 type Channel struct {
 	Name string `json:"name"`
 
+	Backend CommBackend `json:"backend"`
+
+	BrokerHost string `json:"brokerHost"`
```
myungjin, line 34: This openapi file is updated with new entries, but I don't see the api specification updated.

If possible, we should avoid manual changes for openapi files. Manual changes can lead to discrepancy between api specification and the implementation.

myungjin, line 34: Also, per-channel configuration for backend should be optional. WIthout `omitempty`, this line mandates some value in the field.

cmd/controller/app/job/builder.go
```diff
@@ -88,6 +88,10 @@ func (b *JobBuilder) GetTasks(jobSpec *openapi.JobSpec) (
 		return nil, nil, err
 	}
 
+	if err = b.perChannelBackendCheck(); err != nil {
```
myungjin, line 91: I don't see any code to populate per-channel config. 
Where is the following (
https://github.com/cisco-open/flame/pull/373/files#diff-8f0e65135b23fe5484d3a0972cf2b28d79868c2e7dabd76a7bf8f4773da3063bR32-R34) used?

I don't see any example or test-code to use those two variables.

To address the above questions, I think you should create a control plane example for a lib example https://github.com/cisco-open/flame/tree/main/lib/python/flame/examples/hybrid .

myungjin (changes requested):
I left a few comments. Please address my comments. Reach out if there are any questions regarding my comments.

api/design_components.partials.yml
```diff
@@ -159,6 +159,10 @@ Channel:
   properties:
     name:
       type: string
+    backend:
+      $ref: '#/components/schemas/CommBackend'
+    brokerHost:
```
myungjin, line 164: Should this be simply `host` to comply with the design doc you have?

api/design_components.partials.yml
```diff
@@ -217,6 +221,11 @@ Channel:
         value: ["us", "eu"]
       isUnidirectional: false
 
+CommBackend:
```
myungjin, line 224: We already have it defined in `job_components.partials.yml`

take a look at `openapi.yaml.mustache`; all the partials are combined into one file.

So, no need to redefine the same thing here.

cmd/controller/app/job/builder.go
```diff
@@ -188,6 +192,61 @@ func (b *JobBuilder) setup() error {
 	return nil
 }
 
+func (b *JobBuilder) perChannelBackendCheck() error {
```
myungjin, line 195: We shouldn't try to fix any input error. This function tries to check error and fix it. But the function name only suggests that it's checking the error. Anyway, we shouldn't try to fix input error here, IMO.

Also, checking errors in the input should be done somewhere else; for example in preCheck function.

I believe that what we want is to write user-fed per-channel broker and host into config.json (the config used by the lib code) if they are specified in schema.json.

If they (per-channel broker and host) don't exist in the channel section, we simply don't write anything.

I think we should treat it as an error if jobBackend is not specified.

cmd/controller/app/job/builder.go
```diff
@@ -188,6 +192,61 @@ func (b *JobBuilder) setup() error {
...
+			if channel.Backend == "" {
+				jobBackend = openapi.CommBackend(brokers[0].Sort)
+				channel.Backend = openapi.CommBackend(brokers[0].Sort)
+			}
+
+			if channel.BrokerHost == "" {
+				channel.BrokerHost = brokers[0].Host
+			}
+		}
+	}
```
myungjin, line 216: Again, re lines 205-207, let's not complicate things.

We don't need to worry about jobBackend is given or not in order to decide whether populating per-channel configuration is needed or not.

If user provides per-channel broker and host, then we will write them into configj.json. Otherwise, we do nothing.

When per-channel broker and host are given, validating them is a separate issue. checking jobBackend is not empty is also a separate issue.

myungjin (commented):
Left a few comments. You can address those after our meeting scheduled on Monday if you are worried about more drastic changes after the meeting.

pkg/openapi/model_job_spec.go
```diff
@@ -35,7 +35,7 @@ type JobSpec struct {
 	SchemaVersion string      `json:"schemaVersion"`
 	CodeVersion   string      `json:"codeVersion"`
 	Priority      JobPriority `json:"priority,omitempty"`
-	Backend       CommBackend `json:"backend,omitempty"`
+	Backend       CommBackend `json:"backend"`
```
myungjin, line 38: Before manually modifying files under pkg/openapi folder, reconsider if these auto-generated files must be manually updated or not and see if there is a way to avoid that.

If we start to manually update those files, it will become hard to maintain the consistency between api specification and the actual code. Also, later if those files need to be generated again, the manual changes applied to them will not be reflected and it will become hard to track the changes and do them correctly. @openwithcode, this is also FYI.

For this particular case, you need to set backend as required field: https://github.com/cisco-open/flame/blob/main/api/job_components.partials.yml#L53.

cmd/controller/app/job/testdata/hybrid/1.json
```diff
@@ -0,0 +1,56 @@
+{
```
myungjin, line 1: 1.json and 2.json are not configs for the hybrid exmaple. Here are five files you need to refer to.

https://github.com/cisco-open/flame/blob/main/lib/python/flame/examples/hybrid/trainer/config_eu_org1.json
https://github.com/cisco-open/flame/blob/main/lib/python/flame/examples/hybrid/trainer/config_eu_org2.json
https://github.com/cisco-open/flame/blob/main/lib/python/flame/examples/hybrid/trainer/config_us_org1.json
https://github.com/cisco-open/flame/blob/main/lib/python/flame/examples/hybrid/trainer/config_us_org2.json
https://github.com/cisco-open/flame/blob/main/lib/python/flame/examples/hybrid/aggregator/config.json

examples/hybrid/README.md
```diff
@@ -0,0 +1,93 @@
+## Hybrid
```
myungjin, line 1: Looking at the config of the example, this example is not one in https://github.com/cisco-open/flame/tree/main/lib/python/flame/examples/hybrid.

This example can't run the hybrid example that we need to support in this example.

Once the hybrid example is created, update the document correctly. For example, if the example is executed correctly, five agents will be created, not three (https://github.com/cisco-open/flame/pull/373/files#diff-5cc68a51c3891579d0a70d471ca0ac780545df763566eb12d3509d746f294343R67-R69).

myungjin (changes requested):
left a few comments. please consider to fix them.

myungjin (commented):
left a couple of messages.

myungjin (approved):
lgtm

examples/hybrid/dataSpec.json
```diff
@@ -0,0 +1,17 @@
+{
```
myungjin, line 1: The dataSpec's specification is now updated. See https://github.com/cisco-open/flame/blob/main/examples/mnist/dataSpec.json.

myungjin (changes requested):
I spotted another issue. this needs to be addressed.

myungjin:
Since #381 builds on top of this PR, let me close this.

PR #374, opened 2023-03-22
fix: remove non-orchestration mode

openwithcode:
Remove non-orchestration mode from the project as it's not maintained anymore, at least for the medium term.

 The only mode flame will have for now will be the orchestration mode, which means that all the wokers are managed by the flame system.

## Type of Change

- [ ] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist


api/job_components.partials.yml
```diff
@@ -79,46 +79,20 @@ CommBackend:
...
-  type: object
-  additionalProperties:
-    type: integer
-    format: int32
-
-########################################
-# FromSystem
-########################################
-FromSystem:
   type: object
```
myungjin, line 82: This change completely removes model_data_spec.go from the autogenerated code. Also, the produced data structure for it is `map[string]map[string][]string`, which is complicated.

Can you please break this DataSpec into two maps?

DataSpec has role (string) as key and DatasetGroups as value.
DatasetGroups has group (string) as key and list of dataset IDs as value.

We can prevent DataSpec struct from being deleted and keep code readable.

cmd/flamectl/resources/job/job.go
```diff
@@ -62,7 +62,7 @@ func createJobSpec(data []byte, jobFile string) (bool, openapi.JobSpec) {
 	}
 
 	//validate data spec
-	dataSpec := openapi.DataSpec{}
+	dataSpec := make(map[string]map[string][]string)
```
myungjin, line 65: `map[string]map[string][]string` is really unclear. Please refer to the above comment.

pkg/openapi/model_data_spec.go
```diff
@@ -1,49 +0,0 @@
-// Copyright 2022 Cisco Systems, Inc. and its affiliates
```
myungjin, line 1: We should find a way to keep keep this file.

api/job_components.partials.yml
myungjin: How about the following structure?

```
DataSpec:
  description: Dataset specification
  type: array
  items:
    $ref: '#/components/schemas/RoleDatasetGroups'
  example:
    - role: trainer
      datasetGroups:
        eu:
          - 61609290fa724deafdb2a4fa # id of dataset registered in the flame system
          - 61609290fa724deafdb2a4fb # id of dataset registered in the flame system
        us:
          - 61609290fa724deafdb2a4fc # id of dataset registered in the flame system

RoleDatasetGroups:
  description: Dataset specification
  type: object
  properties:
    role:
      type: string
    datasetGroups:
      type: object
      additionalProperties:
        type: array
        items:
          type: string
  example:
    role: trainer
    datasetGroups:
      eu:
        - 61609290fa724deafdb2a4fa # id of dataset registered in the flame system
        - 61609290fa724deafdb2a4fb # id of dataset registered in the flame system
      us:
        - 61609290fa724deafdb2a4fc # id of dataset registered in the flame system
```

pkg/openapi/model_data_spec.go
myungjin: I realized keeping this file is hard. You can ignore my comment above.

cmd/flamectl/resources/job/job.go
myungjin: See the comment at https://github.com/cisco-open/flame/pull/374#discussion_r1148137783.

PR #389, opened 2023-04-03
fix: zero weight upload from trainer in hybrid mode

jaemin-shin:
remove _update_weights() call from _upload_weights() function before calculating the delta weights, as such a function call makes self.prev_weights and self.weights equal, resulting in delta_weights to be all zero

## Description

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/lib/python/flame/mode/hybrid/trainer.py
+++ b/lib/python/flame/mode/hybrid/trainer.py
@@ -109,14 +109,13 @@ class Trainer(DistTrainer):
         # one aggregator is sufficient
         end = channel.one_end()
 
-        self._update_weights()
-
         if not self.can_ring_allreduce():
             # ring was not formed; so, all-reduce didn't take a place.
             # in such a case, each worker should follow the conventional
             # fl procedure by sharing  their local model weights with
             # an aggregator.
             logger.debug("ring was not formed; each shares its weights")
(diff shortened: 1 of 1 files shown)


lib/python/flame/mode/hybrid/trainer.py
```diff
@@ -109,8 +109,6 @@ def _upload_weights(self, tag: str) -> None:
         # one aggregator is sufficient
         end = channel.one_end()
 
-        self._update_weights()
```
myungjin, line 112: This is a nice catch! Thanks!
But this needs some additional change: I think it should be moved to the statement "if not self.can_ring_allreduce():".
If all reduce was executed, this self._update_weights() is problematic because it was already called.
If all reduce was not executed, removing this part means that prev_weights was not updated. To handle this case, I think self._update_weights() needs to be inside the if-statement.

jaemin-shin: I see, you are correct! Should have catched this before I make a PR. Just updated the PR!

myungjin (approved):
lgtm

PR #390, opened 2023-04-04
feat: optimize composer's extensibility

myungjin:
## Description

When a class is inherited, the current compose() function doesn't reuse the tasklets created from the parent class. Therefore, a developer needs to create many of the same tasklets so that they can be chained (hence, composed). This duplicates the code.

To avoid this redundancy, the composer is extended such that the parent's tasklets are retrieved with alias used when they were created in the parent class.

To demonstrate this feature, the compose() method of top_aggregator in asyncfl mode is modified. All other examples still follow the old way of duplicating the code. Those can be revised over time.

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/lib/python/flame/mode/horizontal/asyncfl/top_aggregator.py
+++ b/lib/python/flame/mode/horizontal/asyncfl/top_aggregator.py
@@ -19,15 +19,15 @@ import logging
 import time
 from copy import deepcopy
 
-from ....channel import VAL_CH_STATE_RECV, VAL_CH_STATE_SEND
-from ....common.constants import DeviceType
-from ....common.util import weights_to_device, weights_to_model_device
-from ....optimizer.train_result import TrainResult
-from ...composer import Composer
-from ...message import MessageType
-from ...tasklet import Loop, Tasklet
-from ..top_aggregator import TAG_AGGREGATE, TAG_DISTRIBUTE
-from ..top_aggregator import TopAggregator as SyncTopAgg
+from flame.channel import VAL_CH_STATE_RECV, VAL_CH_STATE_SEND
+from flame.common.constants import DeviceType
+from flame.common.util import weights_to_device, weights_to_model_device
+from flame.mode.composer import CloneComposer
+from flame.mode.horizontal.syncfl.top_aggregator import TAG_AGGREGATE, TAG_DISTRIBUTE
+from flame.mode.horizontal.syncfl.top_aggregator import TopAggregator as SyncTopAgg
+from flame.mode.message import MessageType
+from flame.mode.tasklet import Loop, Tasklet
+from flame.optimizer.train_result import TrainResult
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


elqurio:
I understand the need for extensibility, but I want to make sure I understood this PR. Let me try to explain how I think it works:
- you use `CloneComposer` to wrap the composer to insert new tasklets
- you execute `unlink` on the composer to remove any order of execution inherited from the parent
- then, you give it a new order by creating the rshift chain where inherited tasklets are called by alias, while implemented tasklets are called by variable name (the reason for this is that you don't have another way to reference parent implemented tasklets)

The implementation is more than valid. However I feel like we are introducing complexity that outweighs the extensibility.

It can be boiled down to:
- would you rather check the parent to see what tasklets it has declared and under what alias (and also in what order)
- or would you rather run over the implementation by overriding it (and doing duplicate work)

I think the goal here should have been to give the ability to insert user defined tasklets to the existing chain (if we don't want to change the default ordering). But we kind of get more of the same with more complexity and some piece of mind.

elqurio (dismissed):
Approving per our discussion on the topic.

lib/python/flame/mode/horizontal/asyncfl/top_aggregator.py
```diff
@@ -149,60 +149,52 @@ def _distribute_weights(self, tag: str) -> None:
...
+                >> c.tasklet("train")
+                >> c.tasklet("evaluate")
+                >> c.tasklet("analysis")
+                >> c.tasklet("save_metrics")
+                >> c.tasklet("inc_round")
             )
-            >> task_end_of_training
-            >> task_save_params
-            >> task_save_model
+            >> c.tasklet("inform_eot")
```
jaemin-shin, line 195: how about we use inform_end_of_training instead of inform_eot for the readability of the code?

lib/python/flame/mode/horizontal/coord_syncfl/coordinator.py
```diff
@@ -168,17 +168,17 @@ def compose(self) -> None:
         with Composer() as composer:
             self.composer = composer
 
-            task_check_eot = Tasklet(self.check_eot)
+            task_check_eot = Tasklet("", self.check_eot)
```
jaemin-shin, line 171: here too, in terms of eot

lib/python/flame/mode/horizontal/syncfl/top_aggregator.py
```diff
@@ -267,31 +267,31 @@ def compose(self) -> None:
...
+            task_analysis = Tasklet("analysis", self.run_analysis)
 
-            task_save_metrics = Tasklet(self.save_metrics)
+            task_save_metrics = Tasklet("save_metrics", self.save_metrics)
 
-            task_increment_round = Tasklet(self.increment_round)
+            task_increment_round = Tasklet("inc_round", self.increment_round)
 
-            task_end_of_training = Tasklet(self.inform_end_of_training)
+            task_end_of_training = Tasklet("inform_eot", self.inform_end_of_training)
```
jaemin-shin, line 290: here too, in terms of the eot

jaemin-shin (commented):
I think this is a good direction of optimization, and I also agree to which Lazar pointed out. I have a small concern that we are sacrificing a bit of code readability to avoid duplicated code, but current modification seems to be . Please include me in further discussions.

lib/python/flame/mode/horizontal/syncfl/top_aggregator.py
myungjin: addressed it

lib/python/flame/mode/horizontal/asyncfl/top_aggregator.py
myungjin: addressed it

lib/python/flame/mode/horizontal/coord_syncfl/coordinator.py
myungjin: this is intentionally left empty.

PR #392, opened 2023-04-04
updated feddyn implementation pytorch

GustavBaumgart:
The new implementation matches the algorithm implemented in the paper. This is no longer is similar to the pseudocode they provide.

The new implementation includes an updated communication protocol where all trainers send the dataset size for that round to the aggregator at the beginning of every round. This allows the aggregator to compute adaptive hyperparameters for each trainer individually and then send alpha_adpt along with the global model to begin training for the round.


lib/python/flame/common/constants.py
```diff
@@ -54,3 +54,10 @@ class TrainerState(Enum):
     PRE_TRAIN = 'pre_train'
     DURING_TRAIN = 'during_train'
     POST_TRAIN = 'post_train'
+
+class AggregatorState(Enum):
+    """Enum class for aggregator state."""
+
+    PRE_TRAIN = 'pre_train'
+    DURING_TRAIN = 'during_train'
+    POST_TRAIN = 'post_train'
```
myungjin, line 63: None of DURING_TRAIN, POST_TRAIN seem to be used in this PR. Also, the naming can be improved too.
To be precise, aggregator doesn't have "training"-related state.

Also, TrainerState and AggregatorState are exactly the same. If you want to keep the same set of states (PRE_TRAIN, DURING_TRAIN, POST_TRAIN), why not have only one enum class?

Maybe how about the following?
```
class TrainState(Enum):
    PRE = "pre"
    DURING = "during"
    POST = "post"
```

lib/python/flame/examples/medmnist_feddyn/aggregator/template.json
```diff
@@ -0,0 +1,81 @@
...
+            "name": "param-channel",
+            "pair": [
+                "trainer",
+                "aggregator"
+            ],
+            "funcTags": {
+                "aggregator": [
+                    "distribute",
+                    "aggregate",
+                    "alpha"
```
myungjin, line 35: This function tag is confusing. You may want to use `verb` to indicate it's a function. alpha sounds like a variable / argument; from optimizer alpha is indeed an argument.

lib/python/flame/examples/medmnist_feddyn/aggregator/template.json
```diff
@@ -0,0 +1,81 @@
...
+                ],
+                "trainer": [
+                    "fetch",
+                    "upload",
+                    "dataset"
```
myungjin, line 40: same here. please rename it with a verb.

lib/python/flame/examples/medmnist_feddyn/aggregator/template.json
```diff
@@ -0,0 +1,81 @@
...
+    "registry": {
+        "sort": "dummy",
+        "uri": ""
+    },
+    "selector": {
+        "sort": "default",
+        "kwargs": {}
+    },
+    "optimizer": {
+    "sort": "feddyn",
```
myungjin, line 72: wrong indentation

lib/python/flame/mode/horizontal/feddyn/top_aggregator.py
```diff
@@ -0,0 +1,154 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+"""FedDyn horizontal FL top level aggregator."""
+
+import logging
+
+from ....common.util import (MLFramework, get_ml_framework_in_use,
```
myungjin, line 20: Let's use absolute path instead of relative path.

lib/python/flame/mode/horizontal/feddyn/top_aggregator.py
```diff
@@ -0,0 +1,154 @@
...
+        super().internal_init()
+
+    def get(self, tag: str) -> None:
+        """Get data from remote role(s)."""
+        if tag == TAG_AGGREGATE:
+            self._aggregate_weights(tag)
+        elif tag == TAG_ALPHA:
+            self._save_sizes(tag)
+
+    def _save_sizes(self, tag: str) -> None:
```
myungjin, line 57: perhaps name the function as get_dataset_size() ?

lib/python/flame/mode/horizontal/feddyn/top_aggregator.py
```diff
@@ -0,0 +1,154 @@
...
+        logger.debug("calling _save_sizes")
+        channel = self.cm.get_by_tag(tag)
+        if not channel:
+            return
+
+        self.dataset_sizes = dict()
+        
+        # receive dataset size from all trainers
+        all_ends = channel.all_ends()
```
myungjin, line 66: Why is this needed? why can we not use channel.ends()?

myungjin (commented):
left a couple of comments. some needs an offline discussion.

lib/python/flame/common/constants.py
GustavBaumgart: Yeah, I think I can use the TrainState for both trainer and aggregator updates to the regularizer.

lib/python/flame/examples/medmnist_feddyn/aggregator/template.json
GustavBaumgart: I'm thinking "get_dataset_size"

lib/python/flame/examples/medmnist_feddyn/aggregator/template.json
GustavBaumgart: "send_dataset_size"?

lib/python/flame/examples/medmnist_feddyn/aggregator/template.json
GustavBaumgart: Got it!

lib/python/flame/mode/horizontal/feddyn/top_aggregator.py
GustavBaumgart: channel.ends() only gets the selected ends for that round. The server does need information from all clients (not just those selected) to compute alpha values for the trainers that were selected.

lib/python/flame/examples/medmnist_feddyn/aggregator/template.json
myungjin: How about `uploadDatasetSize`?

lib/python/flame/examples/medmnist_feddyn/aggregator/template.json
myungjin: How about `getDatasetSize`

lib/python/flame/mode/horizontal/coord_syncfl/trainer.py
```diff
@@ -82,7 +82,7 @@ def _send_weights(self, tag: str) -> None:
         channel.await_join()
 
         self._update_weights()
-        self.regularizer.save_state(TrainerState.POST_TRAIN, loc_model=self.model)
+        self.regularizer.save_state(TrainState.POST, loc_model=self.model)
```
myungjin, line 85: why is this needed at coord_syncfl? coord_syncfl is a hierarchical fl. So, I guess we can't use feddyn with coord_syncfl.

lib/python/flame/mode/horizontal/feddyn/top_aggregator.py
```diff
@@ -0,0 +1,154 @@
...
+import logging
+
+from flame.common.util import (MLFramework, get_ml_framework_in_use,
+                            weights_to_device)
+from flame.common.constants import (DeviceType, TrainState)
+from ...composer import Composer
+from ...message import MessageType
+from ...tasklet import Loop, Tasklet
+
+from ..top_aggregator import TopAggregator as BaseTopAggregator
```
myungjin, line 27: Please use the absolute path instead of the relative path.

lib/python/flame/mode/horizontal/syncfl/trainer.py
```diff
@@ -156,7 +157,7 @@ def _send_weights(self, tag: str) -> None:
         end = channel.one_end(VAL_CH_STATE_SEND)
 
         self._update_weights()
-        self.regularizer.save_state(TrainerState.POST_TRAIN, loc_model=self.model)
+        self.regularizer.save_state(TrainState.POST, loc_model=self.model)
```
myungjin, line 160: Now that there is a dedicated trainer.py for feddyn, do we still need this here?

myungjin (commented):
left a few comments

lib/python/flame/mode/horizontal/coord_syncfl/trainer.py
GustavBaumgart: At the time I just saw it was already there, so I just modified the line. Does the new PR remove it? I could also just remove it as well.

GustavBaumgart: This might have happened because this was originally (and still is) in syncfl, even though it is really only used in FedDyn as of now.

lib/python/flame/mode/horizontal/feddyn/top_aggregator.py
GustavBaumgart: Oh I see, will do this for everything in lib/python/flame/mode/horizontal/feddyn

lib/python/flame/mode/horizontal/syncfl/trainer.py
GustavBaumgart: I could move this to the FedDyn trainer only (making the file longer) since it is the only place I use it now. Previously, I thought I might leave it there in case it's used in another optimizer as well, but it might not be worth it since, if removed here, save_state can me removed from other regularizers as well.

lib/python/flame/mode/horizontal/coord_syncfl/trainer.py
GustavBaumgart: Decided to remove this ^

myungjin (approved):
lgtm

lib/python/flame/mode/horizontal/feddyn/top_aggregator.py
```diff
@@ -0,0 +1,153 @@
...
+        with Composer() as composer:
+            self.composer = composer
+
+            task_internal_init = Tasklet("", self.internal_init)
+
+            task_init = Tasklet("", self.initialize)
+
+            task_load_data = Tasklet("", self.load_data)
+
+            task_get_dataset = Tasklet("", self.get, TAG_GET_DATATSET_SIZE)
```
jaemin-shin, line 120: maybe we should name it as "task_get_dataset_size" instead of "task_get_dataset" as it could be misleading?

jaemin-shin (commented):
Overall LGTM!

Some thoughts on sharing data sizes from trainers:
- Could sharing dataset sizes of individual trainers to aggregators be regarded as potential privacy leak? If so, what FedDyn requires for setting up its appropriate alpha value incurs tradeoff between accuracy and privacy, which we may want to consider when further investigating with FedDyn.
- How should dataset sizes of trainers be shared when there are multiple aggregators or hierarchical aggregators? I think we do not need to implement it as of now in this PR, but further evaluating it in our experiments with different configurations could be tricky.

PR #396, opened 2023-04-10
Removal of zip files in examples at the control plane layer

alexandruuBytex:
## Description

ZIP files in a repository makes it hard to do version control, test, debug, etc. Remove the zip files and read the trainer, aggregator and the their variations directly.

## Type of Change

- [ ] Bug Fix
- [ x ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


cmd/flamectl/resources/code/code.go
```diff
@@ -48,6 +49,48 @@ func Create(params Params) error {
 	}
 	url := restapi.CreateURL(params.Endpoint, restapi.CreateDesignCodeEndPoint, uriMap)
 
+	var zipArchivePath = fmt.Sprintf("%s.zip", params.CodePath)
```
myungjin, line 52: I guess the whole part that is implemented here can be put as a separate function instead of doing the implementation in the Create() function. That way readability can be improved.

myungjin (changes requested):
Left several comments. Please also address test errors.

cmd/flamectl/resources/code/code.go
```diff
@@ -17,6 +17,7 @@
 package code
 
 import (
+	"archive/zip"
```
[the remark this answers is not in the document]
alexandruuBytex: We still do feed a zip file. We just create and delete the zip file every time and in this way, we make sure the zip file is always using the latest version of the files that are inside. I tried not creating an archive at all but this requires significantly more changes.
Creating and deleting a zip file is the easiest and more stable solution.

cmd/flamectl/resources/code/code.go
```diff
@@ -144,3 +158,75 @@ func Remove(params Params) error {
 
 	return nil
 }
+
+func CreateZipFile(params Params, pathToSrc, archiveDestination string) error {
```
myungjin, line 162: See if you can leverage some code regarding zip here: https://github.com/cisco-open/flame/blob/main/pkg/util/zip.go.

The implementation may or may not support the arbitrary sub-directory structures. It should be a good starting point. Let's avoid any redundant effort.

myungjin (changes requested):
Left several comments. Please consider to address them. Let's discuss offline if there is any question.

elqurio:
See - https://github.com/cisco-open/flame/pull/408

PR #398, opened 2023-04-19
feat: channel leave functionality for p2p backend

myungjin:
## Description

A functionality to leave a channel is not supported so far. The functionality can be useful when a worker needs to leave and rejoin a channel based on some conditions. To support leave functionality, resource deallocaiton and cleanup of internal state in the backend are implemented.

Note that channel leave functionality for mqtt backend is not complete, which will be done separately in a future PR.

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


lib/python/flame/channel.py
```diff
@@ -234,6 +234,9 @@ async def _get():
             else (None, None)
         )
 
+        # set cleanup ready event
+        self._backend.set_cleanup_ready(end_id)
```
jaemin-shin, line 238: I was initially puzzled why an end is set as cleanup_ready after recv here and after join (in p2p.py), but it seems like an end at a channel is always ready for cleanup except when it's working on a task (like rx). Maybe it would be better to have detailed comments or documentation on the lifecycle of channel and end with cleanup.

jaemin-shin (commented):
Looks like to be a huge update but an important feature for flame. The code looks great overall, but we may find some unexpected cases at future experiments. I would be happy to assist further updates on this in future PRs.

lib/python/flame/backend/mqtt.py
```diff
@@ -478,6 +476,22 @@ async def cleanup(self):
         """Clean up resources in backend."""
         pass
 
+    def set_cleanup_ready_async(self, end_id: str) -> None:
+        """Set cleanup ready event for a given end id.
+
+        This should be called in self._loop thread.
+        This is not yet implemented.
+        """
+        pass
```
elqurio, line 485: Maybe raising NotImplementedError instead of pass?

lib/python/flame/backend/mqtt.py
```diff
@@ -478,6 +476,22 @@ async def cleanup(self):
...
+
+    def set_cleanup_ready(self, end_id: str) -> None:
+        """Set cleanup ready event for a given end id.
+
+        This should be called non self._loop thread.
+        This is not yet implemented.
+        """
+        pass
```
elqurio, line 493: Maybe raising NotImplementedError instead of pass?

elqurio (commented):
Will do the job, certainly. 
One thing I'd like to ask is if this will become the default/desired behavior in general and if it was intended, because it increases the complexity and increases the surface where things can break greatly.

+ minor comment & should AbstractBackend have the set_cleanup_ready abstract method?

lib/python/flame/channel.py
myungjin: added more explanation in the set_cleanup_ready() function.

GustavBaumgart (approved):
lgtm, I appreciate the extra comments

lib/python/flame/backend/p2p.py
```diff
@@ -182,30 +179,95 @@ async def _join_inner():
...
+                if clt_writer is not None:
+                    await clt_writer.notify_end(msg)
+                elif svr_writer is not None:
+                    logger.info("server-side leave channel not implemented")
+                    # TODO: server-side leave channel is not implemented
+                    # we can't use svr_writer because it's context for
+                    # recv_data in BackendServicer; we need to make changes
+                    # for notify_end in BackendServicer
+
+            # save end_ids in the channel as a list
```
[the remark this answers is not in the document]
myungjin: addressed your comment

myungjin:
@elqurio On `minor comment & should AbstractBackend have the set_cleanup_ready abstract method?`

I agree with you. But I think we need to revisit this holistically; we should consider what other methods need to be defined as abstract method by comparing methods in mqtt.py and p2p.py.

PR #399, opened 2023-04-19
feat: additional optimization for channel leave

myungjin:
## Description

Some state was not cleaned up when channel leave happens. There may be chunk manager threads for end ids in a channel. Also, sending heartbeat to meta server for the channel is unnecessary. These two cases are handled here.

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/lib/python/flame/backend/chunk_manager.py
+++ b/lib/python/flame/backend/chunk_manager.py
@@ -17,7 +17,7 @@
 
 import logging
 from datetime import datetime
-from queue import Queue
+from queue import Empty, Queue
 from threading import Thread
 
 from flame.backend.chunk_store import ChunkStore
@@ -29,6 +29,9 @@ logger = logging.getLogger(__name__)
 
 KEY_BACKEND = "backend"
 KEY_CHANNEL = "channel"
+KEY_END_ID = "end_id"
+
+QUEUE_TIMEOUT = 5  # 5 seconds
 
 
 class ChunkThread(Thread):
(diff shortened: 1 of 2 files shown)


lib/python/flame/backend/p2p.py
```diff
@@ -296,10 +302,13 @@ async def _register_channel(self, channel) -> None:
...
 
-            while True:
+            channel_name = channel.name()
+            self.meta_svr_heartbeat[channel_name] = True
+            while self.meta_svr_heartbeat[channel_name]:
                 meta_resp = stub.HeartBeat(meta_info)
                 logger.debug(f"meta_resp from heart beat: {meta_resp}")
-                await asyncio.sleep(HEART_BEAT_DURATION)
+                await asyncio.sleep(META_SVR_HEART_BEAT_DURATION)
+            logger.debug(f"out form meta server heart beat loop on {channel_name}")
```
jaemin-shin, line 311:
```suggestion
            logger.debug(f"out from meta server heart beat loop on {channel_name}")
```

lib/python/flame/backend/chunk_manager.py
```diff
@@ -43,9 +46,15 @@ def __init__(
 
         self._backend = kwargs[KEY_BACKEND]
         self._channel = kwargs[KEY_CHANNEL]
+        self._end_id = kwargs[KEY_END_ID]
 
         self.queue = Queue()
         self.chunk_store = ChunkStore()
+        self._done = False
+
+    def stop(self):
```
jaemin-shin, line 55: Is it enough to just set a flag as True to block the message get? I was wondering if we should free all the items in queue or other attributes of ChunkThread class, as without it, it could be problematic when there's millions of ends leaving frequently? It's quite an extreme case though.

jaemin-shin (commented):
LGTM! Left one suggestion and one comment.

elqurio (approved):
lgtm

lib/python/flame/backend/chunk_manager.py
myungjin: I think this is enough. I am not certain what would happen in case of millions of ends. If you have a specific case in mind, I am happy to chat.

lib/python/flame/backend/p2p.py
myungjin: I will fix this in a separate PR.

PR #401, opened 2023-04-25
feddyn algorithm correction

GustavBaumgart:
The FedDyn algorithm does not send out the same model for training as it uses for evaluation. This is now implemented in this PR.

The labels will be changed to type long before being sent to the loss function in order to avoid NaN values as output.

Lastly, the FedDynRegularizer class now has a .to(device) method in order to transfer the information as needed for subprocesses that use the GPU.


myungjin (approved):
lgtm

PR #418, opened 2023-05-16
feat: CPU metric collection

GustavBaumgart:
## Description

Metric collector now uses a thread to collect information on CPU usage. The data points are (currently) taken every 10 seconds, but this can be adjusted later.

If this layout appears to work well, a similar structure will be used to collect GPU usage metrics, although it will involve more details.

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,13 +16,34 @@
...
+        
+    
+    def cpu_stats(self, freq=10):
+        # as of now, this import is not in the requirements for flame
+        import psutil
+        pid = os.getpid()
+        proc = psutil.Process(pid)
+        
+        while True:
+            self.points['cpu_utilization'].append((time.time(), proc.cpu_percent()))
```
jaemin-shin, line 42: maybe we need better interpretable name than points? how about log or stat_log?

lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,13 +16,34 @@
...
         """Initialize Metric Collector."""
         self.state_dict = dict()
+        self.points = defaultdict(list)
+        cpu_thread = threading.Thread(target=self.cpu_stats)
+        cpu_thread.start()
+        
+    
+    def cpu_stats(self, freq=10):
+        # as of now, this import is not in the requirements for flame
+        import psutil
```
jaemin-shin, line 37: why is it imported here?

lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,13 +16,34 @@
...
+        pid = os.getpid()
+        proc = psutil.Process(pid)
+        
+        while True:
+            self.points['cpu_utilization'].append((time.time(), proc.cpu_percent()))
+            
+            self.points['cpu_memory'].append((time.time(), proc.memory_info().rss))
```
jaemin-shin, line 44: do you know how much overhead are involved in measuring cpu_utilization and cpu_memory in given method?

lib/python/flame/monitor/metric_collector.py
```diff
@@ -38,9 +59,23 @@ def accumulate(self, mtype, alias, value):
         self.state_dict[key] = value + self.state_dict.get(key, 0)
         logger.debug(f"Accumulating metric state_dict[{key}] = {self.state_dict[key]}")
     
+    def process_points(self):
```
jaemin-shin, line 62: I think you could also use more interpretable name here too, "process" itself could be abstract. How about something like "save_points_statistics"?

lib/python/flame/monitor/metric_collector.py
jaemin-shin: also, just out of curiosity, is this supported on other OS than ubuntu?

jaemin-shin (commented):
Overall LGTM, left some minor suggestions and questions.

lib/python/flame/monitor/metric_collector.py
GustavBaumgart: I thought it would be better since psutil is not currently included in the requirements, so it would be easier to toggle it on and off if it is there and not installed. Still haven't added a functionality for how to switch metric collection on and off.

lib/python/flame/monitor/metric_collector.py
GustavBaumgart: Hmm I agree with the idea that there's a better name, but I'm not sure stat_log suggests x,y coordinates. As of now, it does look like other operating systems would be able to run that as well.

lib/python/flame/monitor/metric_collector.py
GustavBaumgart: No idea.

lib/python/flame/monitor/metric_collector.py
jaemin-shin: OS-diagnostic is good. It's still a minor thing, but I am still wondering why that "point" variable have a meaning of x, y coordinate in its name? It could be that if you consider a graph, but more directly, I thought it's just a tuple of time and the measurement value of metric that we want to collect.

lib/python/flame/monitor/metric_collector.py
jaemin-shin: I see. As we do not have separate set of pip requirements to install for new users, I thought it could be much simple to just ask all users to do pip install psutils and import it along with other packages at the beginning of this file. I would like to hear what others think too.

lib/python/flame/monitor/metric_collector.py
jaemin-shin: It's good that psutils are efficient way to measure it, but I think we should have some sense and be able to recommend of which frequency would be appropriate to measure these accurately without overhead from this itself.

PR #420, opened 2023-05-20
feat: utilization/memory usage on CPU and GPU

GustavBaumgart:
## Description

The metric collector now runs two additional threads to track GPU/CPU utilization and memory usage. So far, GPU utilization is still not specific to the process that is running.

Additionally, the metric collector was refactored in order to store samples in a variable stat_log. Names of metrics now use a period to separate the metric type and alias.

setup.py has been updated to include gpustat and psutil.

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,31 +16,123 @@
 """Metric Collector."""
 
 import logging
+import os
+import time
+import threading
+import psutil
+from collections import defaultdict
+from flame.common.util import MLFramework, get_ml_framework_in_use
+from gpustat.nvml import pynvml as N
```
myungjin, line 25: Looks like gpustat's module is not used. If pynvml is the only module used, why don't we install pynvml package directly instead of gpustat?

lib/python/setup.py
```diff
@@ -18,38 +18,40 @@
...
+        "aiostream",
+        "boto3",
+        "cloudpickle",
+        "diskcache",
+        "mlflow==2.0.1",
+        "paho-mqtt",
+        "protobuf==3.19.5",
+        "grpcio==1.51.1",
+        "pydantic",
+        "gpustat",
```
myungjin, line 44: please consider to use pynvml instead of gpustat if gpustat's functionality is not used.

lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,31 +16,123 @@
...
+
+        # GPU monitoring
+        gpu_thread = threading.Thread(target=self.gather_gpu_stats)
+        ml_framework = get_ml_framework_in_use()
+        if ml_framework == MLFramework.PYTORCH:
+            import torch
+
+            if torch.cuda.is_available():
+                gpu_thread.start()
+        elif ml_framework == MLFramework.TENSORFLOW:
```
myungjin, line 48: Is it necessary to check what ml framework is in use for gpu stat monitoring?

I think it may be enough to revise lines 56 and 57.

At N.nvmlInit(), you can check whether htere is an error (e.g., pynvml.nvml.NVMLError_LibraryNotFound).

At dcount = N.nvmlDeviceGetCount(), you can check whether dcount is zero or not.

Then, you may not need to know which ml framework is in use.

lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,31 +16,123 @@
 """Metric Collector."""
 
 import logging
+import os
+import time
+import threading
+import psutil
+from collections import defaultdict
+from flame.common.util import MLFramework, get_ml_framework_in_use
```
myungjin, line 24: It's a good idea to separate standard packages from others (3rd party ones and our own ones).
Perhaps if you apply `black`, it may do reformatting.

myungjin (commented):
left a few comments. Please consider to address them.

lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,31 +16,123 @@
...
         logger.debug(f"Saving state_dict[{key}] = {value}")
-    
+
     def accumulate(self, mtype, alias, value):
         key = self.get_key(mtype, alias)
         self.state_dict[key] = value + self.state_dict.get(key, 0)
         logger.debug(f"Accumulating metric state_dict[{key}] = {self.state_dict[key]}")
-    
+
+    def save_log_statistics(self):
```
jaemin-shin, line 119: can we add more than this? like quartile (25, 50, 75), median, std ? Using numpy array makes lot more easier, as it has its own functions like np.mean, np.min, np.median, np.quartile etc.

jaemin-shin (commented):
Left some comments that could be considered, but generally LGTM.

lib/python/flame/monitor/metric_collector.py
GustavBaumgart: I don't know why, but it looks like because of the way it is packaged in gpustat, gpustat is a lot better at avoiding errors that usually come up when I try to use pynvml directly. I think it tries to use several methods from pynvml but then has backup plans as well.

lib/python/setup.py
GustavBaumgart: Will probably not change this for now.

lib/python/flame/monitor/metric_collector.py
GustavBaumgart: I believe I applied `black`, does it usually do that for you?

lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,31 +16,123 @@
...
+
+            if torch.cuda.is_available():
+                gpu_thread.start()
+        elif ml_framework == MLFramework.TENSORFLOW:
+            import tensorflow
+
+            if tensorflow.config.experimental.list_physical_devices("GPU"):
+                gpu_thread.start()
+
+    def gather_gpu_stats(self, interval=1):
```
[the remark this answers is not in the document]
GustavBaumgart: This, I am not certain of. So far, I haven't run any large-scale example along with the metric collection added in. Maybe I should do that before.

lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,31 +16,123 @@
...
+        dcount = N.nvmlDeviceGetCount()
+
+        while True:
+            for d in range(dcount):
+                handle = N.nvmlDeviceGetHandleByIndex(d)
+
+                # GPU utilization
+                # TO DO: implement metric gathering for process-specific utilization of the GPUs
+                try:
+                    self.stat_log[f"gpu{d}_utilization"].append(
```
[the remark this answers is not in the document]
GustavBaumgart: This would be some additional overhead cost, but I can try.

lib/python/flame/monitor/metric_collector.py
GustavBaumgart: Yeah this was just an initial sort of idea for the stat collection. I can change it to numpy if it's in the requirements.

lib/python/flame/monitor/metric_collector.py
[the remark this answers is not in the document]
GustavBaumgart: Actually, do you think it would be misleading given the way I measure memory usage of the process?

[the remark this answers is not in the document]
jaemin-shin: Memory usage looks fine. My point is that we may need to subtract the utilization from running GPUs that aren't running any Flame processes, which we could at least know. We could do this by post-processing by looking at 0 memory usage, but I would prefer doing this within a same loop, as checking pid wouldn't take much of a overhead.

lib/python/flame/monitor/metric_collector.py
GustavBaumgart: Update: will check overhead and change this

lib/python/flame/monitor/metric_collector.py
```diff
@@ -16,31 +16,131 @@
...
+                except:
+                    logger.debug(f"failed to get GPU utilization of GPU {d}")
+
+                # GPU memory usage of process
+                try:
+                    self.stat_log[f"gpu{d}_memory"].append(
+                        sum(
+                            [
+                                proc.usedGpuMemory
+                                for proc in N.nvmlDeviceGetComputeRunningProcesses(
```
myungjin, line 85: `nvmlDeviceGetComputeRunningProcesses` is called twice (one here and the other in line 72).
I think it can be called only once (e.g., in line 63) and compute both memory and utilization at the same time.
Also, this may be pythonic way, but the code structure look so complicated: embedding for and if statements as an argument of a function (e.g., append) with `sum` function as well.

Can you simplify the code?

myungjin (commented):
left one comment

myungjin (commented):
left one comment

myungjin (approved):
lgtm

myungjin (approved):
lgtm

PR #427, opened 2023-06-01
refactor: monitoring scope adjustment in deployer

myungjin:
## Description

Deployer monitoring functionality does more than it has to do. Deployer only needs to monitor pod. If pod failed, it needs to send that signal to the control plane. It shouldn't update job's status. Currently, in case of pod failure, it directly updates task's status, which is not ideal, either.

TODO: We need a mechanism to inform the control plane of pod failure separately so that the control plane can make a final update on task's status correctly by combining all the information (whether the task itself failed, which caused pod failure or pod itself couldn't run, hence pod failure).

## Type of Change

- [ ] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/cmd/deployer/app/monitor_jobs.go
+++ b/cmd/deployer/app/monitor_jobs.go
@@ -17,9 +17,6 @@
 package app
 
 import (
-	"encoding/json"
-	"fmt"
-	"strconv"
 	"time"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


GustavBaumgart (approved):
lgtm

PR #438, opened 2023-06-20
fix: async FL algorithm to follow FedBuff in aggregation

jaemin-shin:
changed the aggregation process to only aggregate delta weights, scale down when agg goal is achieved, and then add the scaled delta weights to the original weights.

## Description

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/lib/python/flame/mode/horizontal/asyncfl/middle_aggregator.py
+++ b/lib/python/flame/mode/horizontal/asyncfl/middle_aggregator.py
@@ -171,10 +171,6 @@ class MiddleAggregator(SyncMidAgg):
         if not channel:
             return
 
-        if self._agg_goal_weights is None:
-            logger.debug(f"type of weights: {type(self.weights)}")
-            self._agg_goal_weights = deepcopy(self.weights)
-
         # receive local model parameters from a trainer who arrives first
         ends = channel.ends(VAL_CH_STATE_RECV)
         logger.debug(f"ends: {ends}")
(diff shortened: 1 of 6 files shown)


lib/python/flame/mode/horizontal/asyncfl/middle_aggregator.py
```diff
@@ -224,8 +220,10 @@ def _aggregate_weights(self, tag: str) -> None:
         # save global weights before updating it
         self.prev_weights = self.weights
 
-        # set global weights
-        self.weights = self._agg_goal_weights
+        # set global weights, by adding scaled aggregated weights with aggregation goal
+        self.weights = self.optimizer.scale_add_agg_weights(
+            self.weights, self._agg_goal_weights, self._agg_goal
```
myungjin, line 225: Looks like self.weights and self.prev_weights point to the same object.
If so, we need to do a `deepcopy` here.
Can you please confirm if those two point to different objects?

myungjin (commented):
left one comment

lib/python/flame/mode/horizontal/asyncfl/middle_aggregator.py
jaemin-shin: Found these are the same; updated the code to deepcopy weights on prev_weights! Thank you very much for nice catch :)

myungjin (approved):
lgtm

PR #440, opened 2023-07-07
fix: hybrid FL aggregation process without member check

jaemin-shin:
changed the hybrid FL's aggregation to not use member check before ring-reduce anymore, instead use the information given by the top-aggregator. To this end, a new top-aggregator is introduced for Hybrid FL, which sends out the information to the ends about which ends to communicate with during ring-reduce.

## Description

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/lib/python/flame/mode/distributed/trainer.py
+++ b/lib/python/flame/mode/distributed/trainer.py
@@ -121,7 +121,6 @@ class Trainer(Role, metaclass=ABCMeta):
             logger.debug(f"channel not found with tag {tag}")
             return
 
-        self._member_check(channel)
         if not self.can_ring_allreduce():
             # members don't agree, we can't do ring-allreduce
             logger.debug("ring was not formed")
@@ -139,32 +138,22 @@ class Trainer(Role, metaclass=ABCMeta):
         # This method is implemented based on
         # https://github.com/baidu-research/baidu-allreduce/blob/master/collectives.cu
         logger.info("starting ring-allreduce")
-        my_id = channel.get_backend_id()
 
         # do all communication with weights on cpu
         self.weights = weights_to_device(self.weights, DeviceType.CPU)
 
-        rank = self.ends_of_ring.index(my_id)
-        size = len(self.ends_of_ring)
-
         logger.debug(f"weights length = {len(self.weights)}")
-        chunk_size = int(len(self.weights) / size)
-        chunk_sizes = [chunk_size] * size
+        chunk_size = int(len(self.weights) / self.group_size)
+        chunk_sizes = [chunk_size] * self.group_size
 
-        residual = len(self.weights) % size
+        residual = len(self.weights) % self.group_size
         for i in range(residual):
             chunk_sizes[i] += 1
 
-        chunk_ends = [0] * size
+        chunk_ends = [0] * self.group_size
         chunk_ends[0] = chunk_sizes[0]
         for i in range(1, len(chunk_ends)):
             chunk_ends[i] = chunk_sizes[i] + chunk_ends[i - 1]
-
-        recv_from = (rank - 1 + size) % size
-        send_to = (rank + 1) % size
-
-        recvfrom_id = self.ends_of_ring[recv_from]
-        sendto_id = self.ends_of_ring[send_to]
 
         # enable the following "digest" computation lines
         # to check if ring-allreduce work correctly.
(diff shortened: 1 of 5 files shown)


lib/python/flame/mode/distributed/trainer.py
```diff
@@ -121,7 +124,12 @@ def _ring_allreduce(self, tag: str) -> None:
             logger.debug(f"channel not found with tag {tag}")
             return
 
-        self._member_check(channel)
+        if self.is_hybrid:
+            if self.group_identified:
+                self._total_count_check(channel)
+        else:
+            self._member_check(channel)
+
```
GustavBaumgart, line 132: I'd be curious if it would be possible to just redefine `_ring_allreduce` in hybrid/trainer.py instead of using if-else statements here. I think that would be better even if it leads to slightly more code, since it would stay modular.

lib/python/flame/mode/distributed/trainer.py
```diff
@@ -139,14 +147,27 @@ def _do_ring_allreduce(self, channel):
...
+            my_id = channel.get_backend_id()
+
+            rank = self.ends_of_ring.index(my_id)
+            size = len(self.ends_of_ring)
+
+            recv_from = (rank - 1 + size) % size
+            send_to = (rank + 1) % size
+
+            recvfrom_id = self.ends_of_ring[recv_from]
+            sendto_id = self.ends_of_ring[send_to]
```
GustavBaumgart, line 166: Same situation here

lib/python/flame/mode/distributed/trainer.py
```diff
@@ -413,9 +429,33 @@ def _member_check(self, channel) -> None:
         self.ends_of_ring = ends
         return
 
+    def _total_count_check(self, channel) -> None:
```
GustavBaumgart, line 432: If this is only used in hybrid FL, maybe its better to just move it to hybrid/trainer.py as well

lib/python/flame/mode/hybrid/top_aggregator.py
```diff
@@ -0,0 +1,189 @@
...
+        self.end_send_to = {}
+        self.end_receive_from = {}
+        self.end_idx = {}
+        self.end_is_committer = {}
+        self.group_ends = {}
+        self.ends_group = {}
+
+        super().internal_init()
+
+    def parse_hybrid_aggregation_info(self) -> None:
```
GustavBaumgart, line 55: Could we move this below `_distribute_weights`?

lib/python/flame/mode/hybrid/trainer.py
```diff
@@ -22,6 +22,7 @@
 from flame.common.util import weights_to_device, weights_to_model_device
 from flame.mode.composer import Composer
 from flame.mode.distributed.trainer import Trainer as DistTrainer
+from flame.mode.hybrid.top_aggregator import GROUP_UNIDENTIFIED
```
GustavBaumgart, line 25: Not sure what the correct approach is here, but I think that if we are going to use something across different files, it may be best to place it inside of some central location like `constants.py`. Other than that, I think it could also just be redefined here.

lib/python/flame/mode/hybrid/trainer.py
```diff
@@ -85,6 +86,28 @@ def _fetch_weights(self, tag: str) -> None:
         if MessageType.ROUND in msg:
             self._round = msg[MessageType.ROUND]
 
+        if MessageType.HYBRID_METADATA in msg:
+            hybrid_metadata = msg[MessageType.HYBRID_METADATA]
+            self.is_hybrid = True
```
GustavBaumgart, line 91: Why is this being set here?

GustavBaumgart (commented):
Overall looks good. I left a few comments (mostly questioning why changes are in the distributed/trainer.py file).

lib/python/flame/mode/distributed/trainer.py
jaemin-shin: Updated

lib/python/flame/mode/distributed/trainer.py
jaemin-shin: Updated

lib/python/flame/mode/distributed/trainer.py
jaemin-shin: Updated

lib/python/flame/mode/hybrid/top_aggregator.py
jaemin-shin: Updated

lib/python/flame/mode/hybrid/trainer.py
jaemin-shin: Updated

lib/python/flame/mode/hybrid/trainer.py
jaemin-shin: I think this is a bit mixed for flame; for instance, syncfl/trainer.py is using constants from channel.py. asyncfl/top-aggregator is also importing constants from syncfl aggregators. Not sure what is the best design for this. Shall we move all of them to constants.py?

myungjin: You can create a new file, constants.py in the hybrid folder, but I think that's an overkill. I am fine with this since it's only one constant.

myungjin (approved):
lgtm

GustavBaumgart (approved):

PR #666, opened 2025-09-23
Feature: Top 1 perturbation from each trainer

gdadlaney:
This is where we decided to get rid of the flaw in the FwdLLM logic of selecting the top K out of K*10 perturbations at every trainer


myungjin:
@gdadlaney Thank you for your interest and contribution. However, this PR has too many commits. Almost all of them are not yours. We can't accept the PR as is.

